Issue #523, opened 2023-09-28
radiation_map_setup Performance with Bunch Tracking


electronsandstuff:
Hey folks, I am debugging a performance problem related to bunch tracking with radiation fluctuations turned on and thought I would open a github issue related to this. 

I have made a minimal working example of tracking a bunch with CSR and radiation fluctuations which is on the CLASSE file system at the path `/nfs/acc/temp/electronsandstuff/csr-radiation-profiling`. In order to profile the code, I was able to compile bmad + tao with debugging symbols enabled and can then run it with the OSX "Instruments" tool due to grpof not being available on that OS.

The results are below.
<img width="1643" alt="image" src="https://github.com/bmad-sim/bmad-ecosystem/assets/61094787/4f078044-b3a5-43c1-9b3a-b77658549df6">
One of the most expensive parts of the calculation is the subroutine `radiation_map_setup` which takes up about 20% of the programs running time. It seems to be called for every particle in the bunch and at first I thought it was the calculation of the map itself that was causing performance problems. However, looking into the code it seems like there is some caching already there (ie the `ele%rad_map%stale` check). 
<img width="1799" alt="image" src="https://github.com/bmad-sim/bmad-ecosystem/assets/61094787/1f1f8a51-fdfe-4cf5-980e-a5fdeb106417">
Instead, it seems like all of the time is getting spent in memory operations related to the two `ele_struct` objects which are performed regardless of whether it computes the radiation map or not. If there was a way to avoid doing these memory operations for every particle, the tracking could probably be sped up quite a bit. Maybe the stale check could be done in `track1_radiation` instead before the setup code is called?

DavidSagan:
A quick running of the revised code shows a time improvement of around 12%. I will investigate further...

DavidSagan:
Fixed based upon PR #532 implemented.

electronsandstuff:
Thanks! Looks like things scale with the number of particles. Got a pretty solid speedup with the new code for my application.

PR #532, opened 2023-09-30
check for existing data before radiation_map_setup

electronsandstuff:
This is related to the performance issue in  #523. I think I was able to solve it by putting a check before where the `radiation_map_setup` subroutine is called.

Here is the performance on the example in the issue before the change.
```
Benchmark 1: tao -noplot -command exit
  Time (mean ± σ):      2.359 s ±  0.019 s    [User: 2.332 s, System: 0.024 s]
  Range (min … max):    2.336 s …  2.395 s    10 runs
```

Here is the performance with the checks.
```
Benchmark 1: tao -noplot -command exit
  Time (mean ± σ):      1.746 s ±  0.008 s    [User: 1.720 s, System: 0.023 s]
  Range (min … max):    1.736 s …  1.758 s    10 runs
```

It gives about a 25% reduction in running time for this example. Profiling it after the change confirms that the `radiation_map_setup` code takes an insignificant amount of time.
image

### DIFF
--- a/bmad/modules/radiation_mod.f90
+++ b/bmad/modules/radiation_mod.f90
@@ -89,8 +89,10 @@ case (drift$, taylor$, multipole$, ab_multipole$, mask$, marker$);  return
 end select
 
 ! Use stochastic and damp mats
-
-call radiation_map_setup(ele)
+! Only setup if the element produces radiation and there isn't existing non-stale data
+if ((ele%value(l$) /= 0 .and. ele%key /= taylor$) .and. (.not. associated(ele%rad_map) .or. ele%rad_map%stale)) then
+  call radiation_map_setup(ele)
+endif
 
 if (edge == start_edge$) then
   rad_map = ele%rad_map%rm0


ChristopherMayes:
`track1_radiation` is called by `track1`. With OpenMP, this is within a parallelized loop: https://github.com/bmad-sim/bmad-ecosystem/blob/24282a4fac6db1bf119a8295ce0d46de5952da30/bmad/multiparticle/beam_utils.f90#L69

@DavidSagan won't this cause a problem, modifying ele_struct by separate threads?

DavidSagan:
@ChristopherMayes Yes it looks like there is an OpenMP problem here. I will take a look. Also the new test as constructed is invalid since Fortran does not have short-circuiting so testing `associated(ele%rad_map)` and `ele%rad_map%stale` in the same if statement is illegal.

DavidSagan:
I modified the code to avoid the OpenMP problem and added code so that the radiation integral values would be reused if possible. This decreased the run time for the test program by about 15%. (Devel/step17 #534)

DavidSagan:
With the Bmad code change mentioned above the difference with and without the PR code is about 5%. The next PR from me will contain this PR code modified to be valid.

PR #743, opened 2024-01-20
Corrections to the Lynch-Dahl approximation of the multiple scattering sigma in the foil tracking code

sam-grant:
I made some corrections to the calculation of the Lynch-Dahl scattering sigma:

1.  The factor $0.157$ in the expression for `chi2_c`, as defined in the 1991 Lynch-Dahl paper, should be multiplied by $10^{11}$ rather than $10^{7}$. This is to facilitate the conversion of `kg/m^2` to `g/cm^2` and `1/eV^2` to `1/MeV^2`. 
2. The factor $2.007\times10^{-5}$ in the expression for `chi2_alpha` should be multiplied by $10^{12}$, in order to convert to `1/eV^2` to `1/MeV^2`. This makes it $2.007\times10^{7}$ rather than $2.007\times10^{1}$.
3. The expression for `nu` should be `nu = 0.5_rp * omega / (1 - f)` rather than `nu = 0.5_rp * omega / (1.167 * (1 + f))` -- the factor of `1/1.167` comes in during the calculation of `omega` and we need a `-f` rather than a `+f`. 
4. The `sigma` stated in the paper is really `sigma^2`: I added a square root. 

With these changes I get values of sigma consistent with those found using the Rossi-Greisen and Highland approximations, as well as G4beamline (Geant4.10). 

More testing is needed, though!

Thanks,
Sam

### DIFF
--- a/bmad/low_level/track_a_foil.f90
+++ b/bmad/low_level/track_a_foil.f90
@@ -69,7 +69,6 @@ do ns = 1, n_step
       material => ele%foil%material(i)
       z_material = atomic_number(material%species)
       select case (nint(ele%value(scatter_method$)))
-
       case (highland$)
         xx0 = xx0 + r_thick * material%area_density_used / material%radiation_length_used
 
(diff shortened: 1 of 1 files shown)


DavidSagan:
Looks good.

PR #862, opened 2024-03-21
space charge switches correctly exit on error

ndwang:
Previously I saw an issue that track1_bunch_csr is called incorrectly in an element with space_charge_method=fft3d and tracking_method=bmad_standard. Track1_bunch catches this but does not exit. I put in additional err_exit calls to fix it.


DavidSagan:
@ndwang Note: currently Tao sets global_com%exit_on_error = F so this will not affect Tao running and currently there is no way for the User to set global_com%exit_on_error. But setting global_com%exit_on_error can be enabled. Is this something you want?

ndwang:
@DavidSagan Is there a reason why Tao doesn't exit on errors?

DavidSagan:
Yes. If Tao is used within a machine control system, Tao calling err_exit can crash the system which is extremely undesirable. Also if the error was introduced by, for example, a `set` command, some people prefer to correct the error with another `set` command rather than having Tao stop and loosing the state of the simulation.

ChristopherMayes:
Well generally it’s not nice for a program to just quit if it sees an error that’s recoverable.

DavidSagan:
I will modify code to mark particle as lost and return to top level rather than using err_exit.

PR #925, opened 2024-04-19
Automate MacOS package detection, etc.

jsberg-bnl:
I think this will make Bmad build on MacOS, automatically detecting whether you have MacPorts of HomeBrew. Sets CC and CXX for you as well. No more ACC_USE_MACPORTS. I'm going to suggest that a couple MacOS users actually test this before merging...


DavidSagan:
@jsberg-bnl Do we need an environment variable to pick either Homebrew or MacPorts if the both are present?

jsberg-bnl:
It's in there, `BMAD_MAC_PACKAGE`. I didn't put it into `util/dist_prefs`, instead the build will fail with a message to set the variable.

DavidSagan:
Merged. Can you look at https://wiki.classe.cornell.edu/ACC/ACL/OffsiteDoc and make any needed changes? Thanks.

jsberg-bnl:
Will do, hopefully later today.

PR #935, opened 2024-04-30
Remove unnecessary macports/homebrew checking in acc_build_common

jsberg-bnl:
Clean out MacPorts detection from `util/acc_build_common` since it's now done elsewhere. `ACC_USE_MACPORTS` no longer used or needed.


DavidSagan:
I tested with macports.

PR #1051, opened 2024-07-10
ADST show in phasors basis in show spin

mattsignorelli:
I've removed the spin tune output in show taylor and placed it in show spin with the flag `-spin_tune`, also now printed in the phasors basis. I haven't tested if this works yet because of Bmad dependency compile bugs


mattsignorelli:
@DavidSagan  this is tested and ready to merge

PR #1174, opened 2024-08-30
CI: run the pytao test suite with every bmad commit

ken-lauer:
## Changes

* Added a pytao continuous integration test workflow
   * It builds and tests the latest bmad with the latest pytao master
   * It only runs Python 3.12 with shared libraries and MPI off
* Added pytao-focused `conda-setup` action 
* Moved dependency setup to its own composite action, so it could be shared with the pytao workflow
* Adjusted `install_bmad.sh` to use `nproc` instead of just hard-coding `2`. 
   * GitHub actions workers have 4 vCPUs - which resulted in a few minutes shaved off the bmad build in my limited testing.
* Adjust the test suite workflow to build on every commit
   * It's common for collaborators to make feature branches, so it would be nice if forks would run the test suite by default as well

Since it builds bmad and then runs the 600ish tests bundled with pytao, it ends up taking a rather long time (~35 minutes). 

### Future possibilities

* The pytao installation could be used to run additional integration tests. In my opinion:
   * Data files should continue to live in `bmad-doc`/`tao_examples` (or `regression_tests` depending on the purpose)
   * The pytest test suites that utilize those data files should live in the pytao repository
* This could potentially be changed at some point to reuse the build from the primary bmad build/test suite (i.e., the `ci.yml` job) down the line, but I don't have the time for it right now.

### DIFF
--- a/.github/actions/conda-setup/action.yml
+++ b/.github/actions/conda-setup/action.yml
@@ -1,0 +1,34 @@
+name: "conda setup"
+description: "Prepare the pytao conda environment"
+inputs:
+  python-version:
+    description: Python version
+    required: false
+    default: "3.9"
+  environment-file:
+    description: Conda environment file
+    required: true
+    default: "environment.yaml"
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Mambaforge
+      uses: conda-incubator/setup-miniconda@v3
+      with:
+        miniforge-variant: Mambaforge
+        activate-environment: pytao-dev
+        use-mamba: true
+        channels: conda-forge
+        environment-file: ${{ inputs.environment-file }}
+        python-version: ${{ inputs.python-version }}
+
+    - name: Show the installed packages
+      shell: bash -l {0}
+      run: |
+        conda list
+
+    - name: Ensure importability
+      shell: bash -l {0}
+      run: |
+        cd /
+        python -c "import pytao"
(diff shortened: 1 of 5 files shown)


DavidSagan:
@ken-lauer Just so there is clarity: I am waiting for you to indicate that everything is ready for me to look at.

ken-lauer:
@DavidSagan - it will take some time to reduce the test suite duration to a level that's acceptable for your development practices. I'll mark this as draft until I can make some progress in the above-linked PR.

ken-lauer:
This now crashes, as expected from #1241. The reduced Python test suite here doesn't take much additional time beyond the build. Perhaps once the other issue is resolved and the tests pass again, I can mark this as ready for review.

ken-lauer:
I think this is finally ready for your consideration, @DavidSagan.

The pytao test job is about 3 minutes more than the current `master` CI builds, and the normal builds in this PR have been reduced by maybe 2 minutes at times (due to the increase in parallel `make` jobs). I'm not sure how much more it could be reduced, as the job steps have to happen serially (build bmad -> install/test pytao). Caching build dependencies would certainly help, but that comes with its own annoyances.

I reduced the scope of the test suite, aiming to verify the basics on every commit:
1. The Tao shared library remains usable
2. The `pipe` commands remain functional and interpretable by pytao's user-facing methods

DavidSagan:
@ken-lauer Merged. Thanks for your work!

PR #1291, opened 2024-11-14
Added row gap scale factor for curve legends

mattsignorelli:
This PR adds a `curve_legend_row_gap_scale` parameter which can be set for a graph's curve legend to change the spacing between entries. Docs and Tao pipe command etc also all updated.


DavidSagan:
PR #1293 Added qp_legend_struct to hold the curve legend parameters making for a more logical structure.

Issue #1308, opened 2024-11-19
`pipe wall3d_radius` type names should indicate arrays and not scalars


ken-lauer:
**Describe the bug**

For `pipe wall3d_radius`, pytao expects array-valued floats to have a type of `REAL_ARR` and not simply `REAL`. These two lines should be `REAL_ARR` instead:

https://github.com/bmad-sim/bmad-ecosystem/blob/b33b6bd8ed75d81d0b55e9d11381a43aedc538e7/tao/code/tao_pipe_cmd.f90#L7881-L7882

**Minimal Working Example**

```
Tao> reinit tao -clear -init '$ACC_ROOT_DIR/regression_tests/pipe_test/tao.init_wall3d' -noplot
Tao> pipe wall3d_radius 1@0 0.0 0.0
wall_radius;REAL;F;  2.90909090909091E+00
origin;REAL;F;  1.00000000000000E+00;  2.00000000000000E+00;  0.00000000000000E+00
perpendicular;REAL;F;  9.09090909090909E-01; -1.00000000000000E+00;  0.00000000000000E+00
```

DavidSagan:
Fixed.

PR #1310, opened 2024-11-19
CI: regenerate Tao interface commands prior to running PyTao tests

ken-lauer:
* Regenerate Tao interface commands prior to running PyTao tests
   * Ensures that the commands are up-to-date (e.g., new `tao.wall3d_radius`)
   * Ensures that all command examples are added to the test suite

This could have caught the issue in #1308 - assuming that an example was added to the command's documentation string.

### DIFF
--- a/.github/actions/conda-setup/action.yml
+++ b/.github/actions/conda-setup/action.yml
@@ -21,6 +21,7 @@ runs:
         channels: conda-forge
         environment-file: ${{ inputs.environment-file }}
         python-version: ${{ inputs.python-version }}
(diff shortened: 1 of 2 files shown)


ken-lauer:
OK, this is ready for your review @DavidSagan 

You'll note that it now shows the new `wall3d_radius` command added to the Tao interface commands in the Python diff step.
If there were an example in the docstring, it would have caught the type error in #1308

DavidSagan:
Since the PR passes the tests it is good enough for me.

PR #1327, opened 2024-11-26
corrected indices used by z_long kick to account for fft shift

ndwang:
Because srz%fw and the charge distribution are centered around index n2+1 instead of 1, FFT introduces a time shift in srz%w_out. The changes account for it and shift the indices when applying the kick. This avoids modifying srz%w_out itself.


DavidSagan:
@ndwang Looks like there is a conflict that needs to be resolved before merging.

PR #1356, opened 2025-01-08
Add Tao command to set number of OpenMP threads

ken-lauer:
## Background and changes

We were hoping to have a command that allowed for on-demand configuration of OpenMP's number of threads, so I quickly put together the code in this PR.

The code adds a hook for commands of the form `set openmp {param_name} = {value}` (for potential future use of other params).

### Usage

So with the typical `set` syntax:

```
Tao> set openmp num_threads = 3
```

## Question

It'd be good to have this available as part of a global structure that we could get the current value of.

Would you be able to recommend a spot for it, @DavidSagan? Or would you rather take over from this point?


ken-lauer:
@ChristopherMayes mentioned this probably belongs in `global%n_openmp_threads` (or similar). I'll close and redo.

PR #1357, opened 2025-01-09
Add Tao command to set number of OpenMP threads

ken-lauer:
## Background / Changes

Supersedes #1356

On-the-fly configuration of OpenMP's number of threads can be helpful, without having to resort to environment variable changes.

* Moves `s%com%omp_n_threads` to `s%global%n_threads` and updates references (only one which was commented out)
   * `s%com` isn't something users typically modify with `set` or view with `show`, so I think this move is justified
   * (Do we need backward-compatibility for this?)
* Adds `show global` and `pipe global` support to show `n_threads`

### Usage

So with the typical `set` syntax:

```
Tao> set global n_threads = 3
[MESSAGE | 2025-JAN-09 18:06:15] tao_set_openmp_n_threads:
    OpenMP active with number of threads: 3
```

### DIFF
--- a/tao/code/tao_command.f90
+++ b/tao/code/tao_command.f90
@@ -692,7 +692,7 @@ case ('set')
       'universe', 'curve', 'graph', 'beam_init', 'wave', 'plot', 'bmad_com', 'element', 'opti_de_param', &
       'csr_param', 'floor_plan', 'lat_layout', 'geodesic_lm', 'default', 'key', 'particle_start', &
       'plot_page', 'ran_state', 'symbolic_number', 'beam', 'beam_start', 'dynamic_aperture', &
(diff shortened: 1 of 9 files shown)


DavidSagan:
@ken-lauer This looks generally good. One thing that got overlooked is the case when the User sets `n_threads` in an init file. You need to modify the routine `tao_init_global` which is in `tao_init_mod.f90` for this.

ken-lauer:
> @ken-lauer This looks generally good. One thing that got overlooked is the case when the User sets `n_threads` in an init file. You need to modify the routine `tao_init_global` which is in `tao_init_mod.f90` for this.

Thanks - that led me to `reinit tao` which also resets globals. Fixed in [f054de7](https://github.com/bmad-sim/bmad-ecosystem/pull/1357/commits/f054de7f4c53219619a3f55cec3f884e81e0cb8c).

There are some code paths where it may get set multiple times in a row. I switched the message to then instead only show when `n_threads` was changed.

PR #1382, opened 2025-02-07
Tao version update.

DavidSagan:
Tao version update.


DavidSagan:
@jsberg-bnl  @ken-lauer The code and documentation for "quiet" was a mess so I rearranged things. Can you check that this is OK? I am especially concerned that this not adversely affect PyTao users.

Also adding @ChristopherMayes to this.

ken-lauer (approved):
I added support for the new `-quiet` flag options in PyTao https://github.com/bmad-sim/pytao/pull/123 .  I did a brief test with `lat_list` commands that output string lists and found them to be functional in the `-quiet` modes.

The PyTao test suite is passing in this PR, so I believe this is good to go.

PR #1411, opened 2025-02-26
Tao version update.

DavidSagan:
Tao version update.


tao/code/tao_pipe_cmd.f90
```diff
@@ -6925,18 +6945,18 @@ subroutine tao_pipe_cmd (input_str)
 ! Notes
 ! -----
 ! Command syntax:
-!   pipe spin_invariant {flags} {who} {ix_uni}@{ix_branch}|{which}
+!   pipe spin_invariant {-array_out} {who} {ix_uni}@{ix_branch}|{which}
```
ken-lauer: PyTao tests are failing due to this - it should remain `{flags}`, otherwise this turns into the following Python code:

```
cmd = f"pipe spin_invariant {-array_out} {who} {ix_uni}@{ix_branch}|{which}"
```
and `-array_out` is not a valid variable name in the format string.

DavidSagan: Much thanks for the diagnosis. The changes to tao_pipe_cmd comment blocks are for making sure that when the Tao manual gets printed, there are no lines that extend past the right margin.

DavidSagan: @ken-lauer I put in a fix but it looks like the PyTao test is using the unfixed code. Do you understand this?

ken-lauer: PyTao gets the information from the `pipe-interface-commands.json` file - looks like you only updated the Fortran comments.

DavidSagan: Thanks!

PR #1412, opened 2025-02-26
Tao: Expose the beam tracking active element index via C API

ken-lauer:
## Changes

* Adds `s%com%ix_beam_track_active_element`, indicating the current element index being tracked through `tao_beam_track`
* Exposes this value with new C interface exported function `tao_c_get_beam_track_element`.

This pairs with a change in PyTao: https://github.com/bmad-sim/pytao/pull/125

## Why?

* Currently, through the C interface, we can only set `track_type = beam` as a normal command and then wait for the final output text to see what Tao did.
* With this PR, we can spawn a separate thread to poll the status of tracking and generate progress bars, giving the user a better indication as to what's going on

## Screenshots

* Using PyTao in Jupyter Lab:

https://github.com/user-attachments/assets/ed987e51-119b-4da0-96f6-b0c20135a451

* Using PyTao in the terminal:

https://github.com/user-attachments/assets/667e92bf-08b0-49f7-9abb-ad396ab9a3e2

### DIFF
--- a/tao/code/tao_c_interface_mod.f90
+++ b/tao/code/tao_c_interface_mod.f90
@@ -290,4 +290,20 @@ tao_c_interface_com%n_int = 0
 
 end subroutine tao_c_out_io_buffer_reset
 
+!------------------------------------------------------------------------
+!------------------------------------------------------------------------
+!------------------------------------------------------------------------
+!+ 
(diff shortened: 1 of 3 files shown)


DavidSagan:
Very nice!

PR #1425, opened 2025-03-07
CI: add workflow to build/test bmad with just conda

ken-lauer:
## Description

This PR adds a workflow that builds bmad-ecosystem with conda dependencies only, and then runs the regression test suite.

This offers a significant speed-up over the current build procedure, as we can skip building all of the dependencies from source (hdf5, lapack, etc.).

## Speed-up

How much faster is it? It takes about half the time to build and test.

Current build and test workflow - around 16 minutes:
image

New build and test workflow - around 8 minutes:
image

## Changes

* Created a new GitHub Actions workflow that builds bmad (including the regression tests) and runs them using `pytest`
   * This tests on Linux (no MPI, OpenMPI versions) and MacOS ARM64 (no MPI).
   * Modified `.github/scripts/install_bmad.sh` to support building with conda and setting appropriate environment variables
   * Added base conda environment yaml file, shared among the MacOS/Linux build step
* Adds `CONDA_BUILD_TESTS` flag for building regression test binaries with conda (which was not previously configurable)
* Removed some unnecessary syntax in build scripts that I noticed, while debugging local build issues (`if ( [[ ]] ); then` is equivalent to `if [[ ]]; then` in bash, the former just creates an unnecessary subshell.)

## Uh oh

But wait... there's a couple failing tests. They are running correctly, but the results are slightly different. Is this something we should be concerned about?

Ubuntu test failure:

* `photon_test`
  ```
  Regression test local failure: reflect_table-field
     Data from "output.now":     [0.7962371623, 1.5924743247]
     Data from "output.correct": [0.7962371393, 1.5924742786]
     Diff: 4.610000003957282e-08 Diff/Val: 2.8948661835112525e-08
  ```

MacOS test failures:

* `mat6_calc_method_test`

  ```
   Regression test local failure: CRAB_CAVITY1:Tracking:MatrixRow6
     Data from "output.now":     [0.00090788304222579, 9.0246396466931e-05, 0.0, -2.546140381865e-09, 7.6200398421711e-06, 0.99999957183113]
     Data from "output.correct": [0.00090788305575663, 9.0246416763196e-05, 0.0, -2.5393315922217e-09, 7.6200669472254e-06, 0.99999957183113]
     Diff: 2.710505430027915e-11 Diff/Val: 3.5570688151788856e-06
  ```

* `spin_general_test`
  ```
  Regression test local failure: Integral g^3 * b_hat * dn/ddelta
     Data from "output.now":     [2.61104609e-17]
     Data from "output.correct": [2.71191375e-18]
     Diff: 2.339854715e-17 Diff/Val: 1.6236377074503123
  ```
  
## What next?

After addressing the failures above, of course.

If you end up liking this approach, we might choose one or more of the following:

* Keep this in addition to the other workflows
* Use this workflow to replace the existing build/test workflow
* Use this workflow to replace the build procedure in the pytest workflow

### DIFF
--- a/.github/bmad-build-env.yaml
+++ b/.github/bmad-build-env.yaml
@@ -1,0 +1,35 @@
+# conda env create -f environment.yml
+name: bmad-build
+channels:
+  - conda-forge
+dependencies:
+  - compilers
+  - cmake
+  - autoconf
+  - automake
+  - cairo
+  - fftw
+  - fgsl
+  - gnuconfig # [unix]
+  - gsl
+  - hdf5
+  - lapack95
+  - libblas * *openblas
+  - libcblas
+  - liblapack
+  - libtool
+  - m4
+  - make
+  - ncurses
+  - openblas
+  - openmpi
+  - pango
+  - pgplot
+  - pip
+  - pkg-config
+  - python
+  - readline
+  - xorg-libx11
+  - xorg-libxt
+  - xorg-xproto
+  - xraylib
--- a/.github/scripts/install_bmad.sh
+++ b/.github/scripts/install_bmad.sh
@@ -11,7 +11,7 @@ echo "**** Setup Preferences"
 
 echo "Number of processors: $(nproc)"
 
-cat <<EOF >> ./util/dist_prefs
+cat <<EOF >>./util/dist_prefs
 export DIST_F90_REQUEST="gfortran"
 export ACC_PLOT_PACKAGE="pgplot"
 export ACC_PLOT_DISPLAY_TYPE="X"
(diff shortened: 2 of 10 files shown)


ken-lauer:
@DavidSagan this is ready for you to take a look at when you have a chance. There are some notes in the above description about the test failures and a possible path forward - if you think this approach is worthwhile of course!

DavidSagan:
I copied this PR into https://github.com/bmad-sim/bmad-ecosystem/pull/1427 to fix the regression tests. I will close this PR and merge the other PR.

DavidSagan:
@ken-lauer And thanks for your work!

PR #1450, opened 2025-04-05
Fixed infinite loop in Tao when chrom_calc failed.

DavidSagan:
Fixed infinite loop in Tao when chrom_calc failed.


DavidSagan:
@ken-lauer I don't understand why the Conda tests are failing. Can you take a look? Thanks.

DavidSagan:
Since it is only the Mac tests that are failing will merge now and diagnose later...

Issue #1535, opened 2025-05-23
`set bmad_com absolute_time_tracking = T` causes a bad orbit with an `LCAVITY` with more than one step.


ChristopherMayes:
**Describe the bug**
`set bmad_com absolute_time_tracking = T` causes a bad orbit with an `LCAVITY` with more than one step. 

Only, the orbit is fine if `parameter[absolute_time_tracking] = T` is set in the lattice file

**To Reproduce**

`lat.bmad`:
```
no_digested
parameter[geometry] = Open            

parameter[p0c]                      =  5.97820040422691870E+006
parameter[particle]                 = Electron

beginning[beta_a]    =  1.39449126865854395E-001
beginning[alpha_a]   =  9.51137732365840982E-001
beginning[beta_b]    =  3.65359641740668428E-001
beginning[alpha_b]   =  1.99621756285904550E+000

!parameter[absolute_time_tracking] = T
!-------------------------------------------------------

L0AF: Lcavity, 
L =  3.09524400000000011E+000, GRADIENT =  1.87562776115461513E+007,
   RF_FREQUENCY =  2.85600000000000000E+009,
   DS_STEP =  6.19048800000000024E-002

LAT: line = ( L0AF)

use, LAT
```

```
tao -noplot -lat lat.bmad

Tao> sho ele 1
...
Tracking: Electron,   State: Alive
         Position[mm] Momentum[1E-3]        Spin   |
  X:       0.00000000     0.00000000               | t_particle [sec]:        1.03281370E-08  E_tot: 6.40553E+07
  Y:       0.00000000     0.00000000               | t_part-t_ref [sec]:      0.00000000E+00  PC:    6.40532E+07
  Z:      -0.00000000    -0.00000000               | (t_ref-t_part)*Vel [m]:  0.00000000E+00  Beta:  0.999968179
  Particle Phase relative to RF Phase (rad/2pi): 3.96624007E-16


Tao> set bmad_com absolute_time_tracking = T

Tao> sho ele 1
...
Tracking: Electron,   State: Alive
         Position[mm] Momentum[1E-3]        Spin   |
  X:       0.00000000     0.00000000               | t_particle [sec]:        1.03560491E-08  E_tot: 7.17382E+06
  Y:       0.00000000     0.00000000               | t_part-t_ref [sec]:      2.79121048E-11  PC:    7.15560E+06
[...]

PR #1537, opened 2025-05-23
Add expected to fail test related to #1535

ChristopherMayes:
This PR adds a test associated with [bmad-ecosystem issue #1535](https://github.com/bmad-sim/bmad-ecosystem/issues/1535):

- adds a new `regression_tests/pytao` for this and other tests. 
- Add test that ensures that enabling bmad_com.absolute_time_tracking does not alter the initial z position of the first element.
- Marking the test with @pytest.mark.xfail to indicate that it is expected to fail until the underlying bug is resolved.
This preserves the test as a regression safeguard while avoiding false negatives in CI.
- @DavidSagan note that you will need to remove `@pytest.mark.xfail` once this bug is fixed.


regression_tests/pytao/lcavity/test_lcavity.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Regression test for issue #1535:
+    Ensure that enabling `absolute_time_tracking` does not alter the initial 
+    longitudinal position `z` of element 1 in the lattice.
+    
+    This checks that switching `bmad_com.absolute_time_tracking` from False (default) 
+    to True does not affect the reported `z` position of the first element.
+    """
+    # Initialize Tao without plotting
+    tao = Tao(lattice_file='lat.bmad', noplot=True)
```
ken-lauer: A blanket `xfail` means that if `Tao` can't find this lattice file, the test will xfail for a reason that you didn't expect.

`Tao(lattice_file="lat.bmad")` needs a full path for the lattice file. See how the snapshot test handles this in the level above:
https://github.com/bmad-sim/bmad-ecosystem/blob/bebdfcd0898c9073b8b3074c055534979f669ead/regression_tests/test_snapshots.py#L12C1-L12C46

You should consider doing this within the function: https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

```python
if not np.isclose(z_before, 0):
    pytest.xfail(f"Expected z=0 before setting absolute_time_tracking, got {z_before}")
```

ChristopherMayes: Thanks, see it now.

ChristopherMayes:
@ken-lauer is there simpler method to locate the lat file in such a test?

ken-lauer:
The preferable method - clearly I'm biased here - is to follow the pattern suggested above.

Add a `conftest.py` with `PYTAO_TEST_ROOT` or similar and then use `Tao(lattice_file=PYTAO_TEST_ROOT / "lcavity" / "lcavity.bmad")` as it's explicit and easy to understand where the file is expected to be. Using `pathlib.Path` is almost always preferable to `os.path`.

MichaelEhrlichman:
Looks like a good application of the suggested `Tao.from_lattice_contents(lat: str)` functionality.

ChristopherMayes:
> The preferable method - clearly I'm biased here - is to follow the pattern suggested above.
> 
> Add a `conftest.py` with `PYTAO_TEST_ROOT` or similar and then use `Tao(lattice_file=PYTAO_TEST_ROOT / "lcavity" / "lcavity.bmad")` as it's explicit and easy to understand where the file is expected to be. Using `pathlib.Path` is almost always preferable to `os.path`.

I changed to use `pathlib.path`. I don't want to have to make another file (conftest.py) here, because that becomes more overhead for a user to worry about when making a test. Having a feature like https://github.com/bmad-sim/pytao/issues/136 will make things simpler.

MichaelEhrlichman (approved):

DavidSagan (approved):

DavidSagan:
@ken-lauer Do you approve?

ken-lauer:
> I don't want to have to make another file (conftest.py) here, because that becomes more overhead for a user to worry about when making a test. 

I disagree; reusable tools and settings make things simpler.

ken-lauer (approved):

PR #1565, opened 2025-06-05
REF/FIX: retain n_threads after 'reinit tao -clear'

ken-lauer:
This PR makes `s%global%n_threads` persist even after a `reinit tao -clear`.

`s%global%n_threads` could be considered a bit of a special case; resetting it may end up surprising the user and result in unexpectedly slow simulations.

### DIFF
--- a/tao/code/tao_parse_command_args.f90
+++ b/tao/code/tao_parse_command_args.f90
@@ -25,7 +25,7 @@ character(200) :: cmd_words(12)
 character(80) arg0, arg1, base, switch
 character(*), parameter :: r_name = 'tao_parse_command_args'
 
-integer n_arg, i_arg, ix
+integer n_arg, i_arg, ix, old_n_threads
 logical error, negate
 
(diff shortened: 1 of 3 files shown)


DavidSagan:
@ken-lauer Should the user even have the ability to set n_threads? There is no documentation on n_threads in the Tao manual.

ken-lauer:
I thought we agreed on the user's ability to set `n_threads` in the previous PR?
https://github.com/bmad-sim/bmad-ecosystem/pull/1357

Would you like me to put something in the manual for this PR?

DavidSagan:
@ken-lauer Sorry I had forgotten that conversation. Yes if you could put something in the manual about this it would be appreciated.

DavidSagan:
@ken-lauer Is this ready to be taken out of draft status?

ken-lauer:
I think this should be good to go, however I haven't been able to test locally due to the recent CMake issue.

Issue #1555, opened 2025-06-03
Conda Builds Failing


electronsandstuff:
Just making a note here that conda builds appear to be failing since 5/27/2025.

<img width="1350" alt="Image" src="https://github.com/user-attachments/assets/41f5a290-f29f-46d9-acf1-aa42d9b96404" />

ken-lauer:
[osx arm64 openmpi](https://github.com/conda-forge/bmad-feedstock/pull/563/checks?check_run_id=43364111100) build failures are causing the workflow to fail after the merge of #1538:

```
2025-06-03T07:41:22.9398080Z -- HDF5 Fortran compiler wrapper is unable to compile a minimal HDF5 program.
2025-06-03T07:41:23.3491410Z -- Found HDF5: $PREFIX/lib/libhdf5_fortran.dylib;$PREFIX/lib/libhdf5.dylib (found version "1.14.3") found components: Fortran HL
2025-06-03T07:41:23.3578360Z SHARED DEPS          : gomp
2025-06-03T07:41:23.3624500Z SHARED LINKER FLAGS  : -fopenmp -Wl,-headerpad_max_install_names -Wl,-dead_strip_dylibs -Wl,-rpath,/Users/runner/miniforge3/conda-bld/bmad_1748935962326/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib -L/Users/runner/miniforge3/conda-bld/bmad_1748935962326/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib
2025-06-03T07:41:23.3763760Z 
2025-06-03T07:41:23.3820640Z -- Configuring done (37.0s)
2025-06-03T07:41:23.4210410Z CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
2025-06-03T07:41:23.4259130Z Please set them or make sure they are set and tested correctly in the CMake files:
2025-06-03T07:41:23.4275750Z HDF5_hdf5_hl_fortran_LIBRARY
2025-06-03T07:41:23.4277310Z     linked by target "forest-shared" in directory /Users/runner/miniforge3/conda-bld/bmad_1748935962326/work/forest
[...]

PR #1566, opened 2025-06-06
MAINT: potential fix for hdf5 fortran HL on arm64+openmpi+conda

ken-lauer:
I'm no CMake expert, but this patch appears to work (ref #1555) on conda-forge: https://github.com/conda-forge/bmad-feedstock/pull/564/checks

### DIFF
--- a/util/Master.cmake
+++ b/util/Master.cmake
@@ -502,8 +502,9 @@ foreach(h5dir ${HDF5_Fortran_INCLUDE_DIRS})
   endif()
 endforeach()
 if (HDF5_FOUND)
+  list(FILTER HDF5_Fortran_HL_LIBRARIES EXCLUDE REGEX "-NOTFOUND$")
+  list(FILTER HDF5_Fortran_LIBRARIES EXCLUDE REGEX "-NOTFOUND$")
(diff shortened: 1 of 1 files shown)


jsberg-bnl:
How about just doing a one-liner,
```
list(FILTER SHARED_LINK_LIBS EXCLUDE REGEX "-NOTFOUND$")
```
in the original `if` block right after `set(SHARED_LINK_LIBS ...`

DavidSagan:
@jsberg-bnl Should I merge?

DavidSagan:
Will merge since people are requesting a Conda fix.

DavidSagan:
@ken-lauer Can you take this out of draft mode? Thanks.

jsberg-bnl:
Re-thinking a bit, maybe two lines fixing `HDF5_Fortran_HL_LIBRARIES` and `HDF5_Fortran_LIBRARIES` individually, since FindHDF5.cmake messes those up when the wrapper is busted. Prevents hiding other issues. But existing should function fine as well.

DavidSagan:
```
Re-thinking a bit, maybe two lines fixing HDF5_Fortran_HL_LIBRARIES and HDF5_Fortran_LIBRARIES individually, 
since FindHDF5.cmake messes those up when the wrapper is busted. Prevents hiding other issues. 
But existing should function fine as well.
```
@ken-lauer Do you want to make the change  and then I will merge?

PR #1559, opened 2025-06-04
deta/dpz correct code.

DavidSagan:

PR #1595, opened 2025-06-25
Now bmad_com[normalize_twiss] default is False.

DavidSagan:


DavidSagan:
@ken-lauer I'm not seeing a failure when I run pytest on my linux machine. Do you understand this?

ken-lauer:
(I misread your initial comment, but I'll leave this anyway)
It's important to look at the error message when the tests fail; they are verbose and explain what's going on.

The situation is the same as in the previous time this failed: https://github.com/bmad-sim/bmad-ecosystem/pull/1559

The relative tolerance likely should be adjusted (based on your knowledge of the formulas/how the calculations work) to be the most that you'd expect that the values would vary between bmad-ecosystem releases. The goal is to check that there weren't major bugs introduced that cause wild fluctuations in standard outputs.

(As to why they might pass on your machine and fail on GitHub)
There may be minor differences between results on different machines due to different libraries/architectures/etc. That's another variable in how you might approach setting the tolerances appropriately.

DavidSagan:
@ken-lauer I did check. The deviations between actual and desired are much larger than I would expect given that the test passes on my machine. I will set the tolerance lower but this is concerning.

ken-lauer:
I see - if that's the case, then the test suite is doing its job in alerting you about multi-platform issues early. 

What could there be in your algorithm/setting changes that might even have a slight effect on it?

DavidSagan:
I know what I did to change the numbers but I have no idea as to why the platform-to-platform shift is so large.

ChristopherMayes:
@DavidSagan what was the motivation for this? Could you add a description?

DavidSagan:
The motivation for this was to get Twiss plots for lcavity and e_gun elements to agree with the standard thick element calculation. There is now a section in the Bmad manual on normalizing the matrix for the Twiss calculation.

Issue #1619, opened 2025-07-16
Multipass `lcavity`, No Energy Gain in `bmad-standard` Tracking with `n_rf_steps` > 0


electronsandstuff:
**Describe the bug**
Related to #1610, when using `n_rf_steps` with multipass cavities, the reference energy appears unchanged by `lcavity` elements.

**Minimal Working Example**
The following lattice shows this behavior.
```
no_digested
beginning[beta_a] = 10
beginning[beta_b] = 10
beginning[e_tot] = 10e6
parameter[geometry] = open
parameter[absolute_time_tracking] = T

p: patch
cav: Lcavity, voltage = 1e6, L = 1, rf_frequency = 1e9
linac: line[multipass] = (cav)

lat: line = (linac, p, linac)
use, lat

Lcavity::*[n_rf_steps] = 1024
```

**To Reproduce**
1. Load the lattice in `tao`
2. Run `sho ele cav\1`
3. Note that the start/end energy of the cavity is the same
4. Comment out the final line setting `n_rf_steps`
5. Load the lattice and rerun command
6. Note that energy is increased by cavity voltage.

**Compiler and Operating system**
Conda Bmad version 20250709, MacOS

electronsandstuff:
I have also made a regression test that will fail on the issue and added it in the linked PR.

DavidSagan:
@electronsandstuff This looks like a bug that was recently fixed so I suspect you were not using the latest code. Please check.

electronsandstuff:
OK, updating did resolve the issue in the MWE. I had to do a little fighting with conda to update to the latest version which is likely why it wasn't pulled in before.

Unfortunately, `tao` now crashes due to a segfault in my application. I will work on preparing a report in a separate issue from this one.

PR #1620, opened 2025-07-16
Add Regression Test for Multipass `lcavity` Issue

electronsandstuff:
This PR adds a regression test that fails on the issue reported in #1619.


DavidSagan:
@electronsandstuff Looks like there is a problem with the file location which is causing test failure:
```
==================================== ERRORS ====================================
__________ ERROR collecting pytao/lcavity_n_rf_steps/test_lcavity.py ___________
import file mismatch:
imported module 'test_lcavity' has this __file__ attribute:
  /home/runner/work/bmad-ecosystem/bmad-ecosystem/regression_tests/pytao/lcavity/test_lcavity.py
which is not the same as the test file we want to collect:
  /home/runner/work/bmad-ecosystem/bmad-ecosystem/regression_tests/pytao/lcavity_n_rf_steps/test_lcavity.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

electronsandstuff:
I will look into it. It seems like it is confused because this test has the same filename as the other pytest test.

electronsandstuff:
Here is the issue.

https://stackoverflow.com/questions/53918088/import-file-mismatch-in-pytest

Pytest uses old python imports and fails on files with the same name unless the file `__init__.py` exists through the file structure so it is seen as a python package.  This is why the test worked locally from the test directory, but fails on CI. Details in the link for those who are interested. 

Our options are the following:
- Add `__init__.py` to the test directory (and I think the subdirectories). We could also refactor so the python files are in a top level directory and pull lattices from a data directory. This is how I typivally structure things anyways.
- Change the filename and just require no duplicate names.
- The Stack Overflow link says you can force pytest to use the newer python import system. I am unsure if this comes with any side effects.

@ChristopherMayes, any opinions?

ChristopherMayes:
@electronsandstuff I'd try to force the new import system to avoid having to add boilerplate `__init__.py` all over the place.

electronsandstuff:
I tried out the suggested CLI option, but a different import issue related to `pytao` showed up. Because of this, I decided to rename the test instead and add a note to the directory explaining that files must have unique names.

Issue #1693, opened 2025-08-15
`PATCH[T_OFFSET]` Affects Second Pass Energy with `n_rf_steps>0`


electronsandstuff:
**Describe the bug**
In a simple lattice with multipass cavities and a patch element, changing `T_OFFSET` affects final energy at the 1e-5/1e-4 level. This is present in the new cavity model (ie `bmad-standard` tracking with `n_rf_steps>0`).

Related issue: https://github.com/bmad-sim/bmad-ecosystem/issues/1610

**Minimal Working Example**
The following lattice exhibits the issue. A unit test is added in #1694 
```
no_digested
beginning[beta_a] = 10
beginning[beta_b] = 10
beginning[e_tot] = 10e6
parameter[geometry] = open
parameter[absolute_time_tracking] = T

p: patch
cav: Lcavity, voltage = 1e6, L = 1, rf_frequency = 1e9, n_rf_steps=256
linac: line[multipass] = (cav)

lat: line = (linac, p, linac)
use, lat
```

**To Reproduce**
1. Open lattice in tao
2. Observe final energy (`sho value ele::end[orbit.energy]`)
3. Modify patch `T_OFFSET` (`set ele p t_offset = 0.3e-9`)
4. Note final energy has changed (`sho value ele::end[orbit.energy]`)

**Expected behavior**
I would expect no change in final energy/timing.

**Compiler and Operating system**
Bmad 20250811.2 from conda on MacOS

ChristopherMayes:
This shows the variation in relative orbit energy error with patch `t_offset`:

```python
from pytao import Tao
import numpy as np
import matplotlib.pyplot as plt

tao = Tao(lattice_file='lat.bmad', noplot=True)

frf = tao.ele_gen_attribs('cav')['RF_FREQUENCY']

def ptot(orbit):
    return (1+orbit['pz']) * orbit['p0c']

def ptot_from(t_offset):
    tao.cmd(f"set ele p t_offset = {t_offset}")
    return ptot(tao.ele_orbit(r"cav\2"))    

t_offsets = np.linspace(0, 1/frf, 100)
ps = np.array([ptot_from(t_offset) for t_offset in t_offsets])

rel_error = ps/ps[0] -1

plt.plot(frf * t_offsets, rel_error*1e6)
plt.xlabel(r'$f_\text{rf} * \text{t_offset}$')
plt.ylabel(r'$p/p_0 -1$ ($10^{-6}$)')    
```

[...]

PR #1694, opened 2025-08-15
Add Test of Second Pass Timing with Patch Element and `n_rf_steps` > 0

electronsandstuff:
Regression test for issue #1693


DavidSagan:
@electronsandstuff If you merge the latest code into this PR this may fix things.

PR #1729, opened 2025-09-22
Fix expression evaluation problem.

DavidSagan:


DavidSagan:
@ken-lauer @ChristopherMayes There are now two more `pipe` subcommands called `lord_control` and `slave_control`. The `lord_control` test is failing and I don't understand why since the output from Tao looks OK.

ChristopherMayes:
These need new custom parsers. I'm adding those in https://github.com/bmad-sim/pytao/pull/144

Issue #1762, opened 2025-10-15
Incorrect Energy Change in Cavity at Low Energy in New Cavity Model


electronsandstuff:
**Describe the bug**
For an `LCAVITY` element with a non-ultrarelativistic beam, energy change seems to no longer be calculated correctly starting in Bmad version 20250915.1 (introduction of `n_rf_steps=10` as default value).

In the example lattice below, the cavity shows no energy change in Bmad 20250915.1
```
Tao> sho ele l0af

Attribute values [Only non-zero values shown]:
...
   52  E_TOT_START                 =  8.0991512E+05 eV           DELTA_E                     =  0.0000000E+00 eV

```

Here is the same lattice in Bmad version 20250915.0.
```
Tao> sho ele l0af

Attribute values [Only non-zero values shown]:
...
   52  E_TOT_START                 =  8.0991512E+05 eV           DELTA_E                     =  9.1677166E+04 eV
```

**Minimal Working Example**
I have prepared a working example which is available in PR #1763. Copying here for ease of use.
```
no_digested
parameter[geometry] = Open            

parameter[p0c]                      =  6.28365E+05
parameter[particle]                 = Electron

beginning[beta_a]    =  1.39449126865854395E001
beginning[alpha_a]   =  9.51137732365840982E-002
beginning[beta_b]    =  3.65359641740668428E001
beginning[alpha_b]   =  1.99621756285904550E-002

!parameter[absolute_time_tracking] = T
!-------------------------------------------------------

L0AF: Lcavity,  L =  0.4, GRADIENT =  1e6,
   RF_FREQUENCY =  0.900000000000000E+009,
   phi0=-0.25, mat6_calc_method=tracking,
   tracking_method=Time_Runge_Kutta

LAT: line = ( L0AF)

use, LAT
```

**To Reproduce**
Open the lattice file with `tao`. Run `sho ele l0af` and check the key `DELTA_E`.

**Expected behavior**
Energy change should show up when tracking through cavity at low energy.

**Compiler and Operating system**
MacOS Bmad from conda.

DavidSagan:
This is due to the change in how the change in reference energy through an lcavity is computed. The current calculation does not do tracking but rather uses the `voltage` and `phi0` settings. A problem I just discovered and fixed was in the calculation of `phi0_autoscale` which was not making sure that a particle entering on the reference orbit leaves on the reference orbit.

PR #1763, opened 2025-10-15
Add Regression Test for Low Energy Tracking Through Buncher Cavity

electronsandstuff:
This PR adds an accompanying regression test and minimum working example for #1762.


DavidSagan:
@electronsandstuff Looks like there are conflicts there that need to be resolved. Note: I did just add a test for this on my side.

electronsandstuff:
OK, resolved if you want to add this one too.

DavidSagan:
@electronsandstuff Looks like now there are failing tests...

electronsandstuff:
I am checking on my end and there still seems to be an issue in the cavity. That's why the new unit test is failing. Here is Bmad 20251016.0 on the test lattice.
image
Here is Bmad 20250915.0
image
I think something is still going on with the energy calculations, but I also don't have permissions to re-open the issue.

DavidSagan:
I assume by "issue" you mean #1762. See my comment at the end of this issue. The fact that there is a difference now as compared to 20250915.0 is to be expected.

electronsandstuff:
Sorry if I am misunderstanding something, but I am still confused.

I know that `E_tot` is the reference energy, but since the lattice doesn't have any pz orbit in either Bmad version, it is also the energy of the bunch.

In the older Bmad version, the code says that the bunch gains 100keV through the cavity. In the latest Bmad, the bunch's energy doesn't change (no change in `E_tot` or orbit pz). Outside of convergence issues, I would expect there to be a consistent value for the beam's energy after the cavity. Maybe I am calculating it incorrectly or missing something?

DavidSagan:
Yes there has been a change and the results are not consistent. What the code did in the past was not what a User would expect so this has changed.

DavidSagan:
If you want we can zoom and I can explain in detail.

electronsandstuff:
OK, thanks. I will reach out to you over slack. In the meantime it sounds like this test isn't correct and I can close out the PR.

PR #1866, opened 2026-01-07
Fix interface argument names, some `optional` annotations

ken-lauer:
Changes:
* Fixes some bmad routine interface argument names to match the underlying implementation source
* Fixes some documentation argument `optional` annotations which were inaccurately marked as `optional`
* Removes a number of missing routines from the interface definitions - I assume these were deleted from the codebase at some point in the past

For context, this information is parsed and used to generate the new C++ and Python bindings.

### DIFF
--- a/bmad/custom/em_field_custom.f90
+++ b/bmad/custom/em_field_custom.f90
@@ -41,7 +41,7 @@
 !
 ! Output:
 !   field    -- Em_field_struct: Structure hoding the field values.
-!   err_flag -- Logical, optional: Set true if there is an error. False otherwise.
+!   err_flag -- Logical: Set true if there is an error. False otherwise.
 !-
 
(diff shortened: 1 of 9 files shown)


bmad/custom/em_field_custom.f90
```diff
@@ -41,7 +41,7 @@
 !
 ! Output:
 !   field    -- Em_field_struct: Structure hoding the field values.
-!   err_flag -- Logical, optional: Set true if there is an error. False otherwise.
+!   err_flag -- Logical: Set true if there is an error. False otherwise.
```
DavidSagan, line 44: `err_flag` is marked as optional in the code so this change is not correct.

DavidSagan: @ken-lauer I will merge this PR and deal with the doc later...

PR #1882, opened 2026-01-17
Refactor mark_patch_regions algorithm for performance

ken-lauer:
Note: the new function generates the same results for our lattice, but I imagine I could still be missing something here with respect to the inner workings. Any thoughts?

### Background

In large lattices, Tao can slow to a crawl when attempting to determine if there are patch elements in wall segments.

This happens due to `mark_patch_regions`, where:
* For each of the wall3d sections, it finds the nearest elements
* Then it uses the (relatively) heavy `pointer_to_next_ele` to search through for patch elements

Special handling of the wrap-around case for closed geometries makes this a bit more complicated and confusing to read as well.

### Changes

This PR proposes flipping things around a bit:
1. Finding all of the patch elements in advance - just once
2. For each patch element, check if the wall segments fall within the element's `s` position

This significantly reduces the computational complexity, resulting in a savings of almost 4 seconds for one of our specific lattices - it actually almost halves the Tao startup time.

### DIFF
--- a/bmad/modules/wall3d_mod.f90
+++ b/bmad/modules/wall3d_mod.f90
@@ -1540,54 +1540,59 @@ subroutine mark_patch_regions (branch)
 type (branch_struct), target :: branch
 type (wall3d_struct), pointer :: wall
 type (ele_struct), pointer :: ele
-integer i, i0, iw, ie0, ie1, n_sec
-logical first
+integer i, j, iw, n_sec, n_patch, ix1, ix2
+integer, allocatable :: patch_ixs(:)
+real(rp) s_min, s_max
 
 !
+
+allocate (patch_ixs(branch%n_ele_track))
+n_patch = 0
+do i = 1, branch%n_ele_track
+  if (branch%ele(i)%key == patch$) then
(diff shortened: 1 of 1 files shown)


DavidSagan:
@ken-lauer Could you open an issue and post a problem lattice? I want to do some checking. Thanks.

ChristopherMayes:
Here is a `lat.bmad` that calls a detailed `wall.bmad` that is slow to parse. Parsing without the wall is very fast. 

```
time tao -lat lat.bmad -noplot -command quit
[MESSAGE | 2026-JAN-16 19:06:03] tao_init:
    OpenMP active with number of threads: 10
[MESSAGE | 2026-JAN-16 19:06:04] bmad_parser:
    Lattice parse time(min): 0.00
[INFO 2026-JAN-16 19:06:04] tao_command:
    Stopping.

0.71s user 0.04s system 99% cpu 0.756 total
```

[Archive.zip](https://github.com/user-attachments/files/24684396/Archive.zip)

DavidSagan:
Simplified code in #1886 so closing this PR.

Issue #1889, opened 2026-01-21
Question: `apply_element_edge_kick` using `em_field_calc` with `calc_potential`


ken-lauer:
During profiling of loading a large lattice with Runge-Kutta tracking, I noticed that roughly 40% of the time loading the lattice was spent calculating electrical/magnetic potentials in [`em_field_calc`](https://github.com/bmad-sim/bmad-ecosystem/blob/2349dee9bb5837f41df8315a04806b94df63520a/bmad/code/em_field_calc.f90#L46).

I traced this back to [`apply_element_edge_kick`](https://github.com/bmad-sim/bmad-ecosystem/blob/2349dee9bb5837f41df8315a04806b94df63520a/bmad/low_level/apply_element_edge_kick.f90#L131) which explicitly enables `calc_potential` (the last `.true.` parameter).

@ChristopherMayes indicated that he thinks this is perhaps unintentional - or maybe even unnecessary.

If it's required, would you please clarify for us why that's the case?

Regardless, I have a branch that would significantly speed up the `calc_potential=.true.` case, avoiding nested function trampolining (at least with gfortran). I'll propose that later with some more details.

DavidSagan:
Sorry I missed the question. I do not see anyplace in apply_element_edge_kick that deals with the potential.

ken-lauer:
`em_field_calc`: https://github.com/bmad-sim/bmad-ecosystem/blob/2349dee9bb5837f41df8315a04806b94df63520a/bmad/code/em_field_calc.f90#L46-L47
`apply_element_edge_kick` call:

https://github.com/bmad-sim/bmad-ecosystem/blob/2349dee9bb5837f41df8315a04806b94df63520a/bmad/low_level/apply_element_edge_kick.f90#L131

Line 131, after `err_flag` is `.true.` - which corresponds to `calc_potential`.

DavidSagan:
Are you free at 2:30 today? (use my zoom).

On Jan 27, 2026, at 12:10 PM, Ken Lauer ***@***.***> wrote:

[https://avatars.githubusercontent.com/u/152229072?s=20&v=4]ken-lauer left a comment (bmad-sim/bmad-ecosystem#1889)<https://github.com/bmad-sim/bmad-ecosystem/issues/1889#issuecomment-3806422390>

[...]

PR #1891, opened 2026-01-22
`em_field_calc`: move up `rb_field` to avoid trampoline of internal procedure

ken-lauer:
(Related to #1889 but does not close it)

In short: a massive performance penalty on MacOS with RK tracking can be avoided by moving around where `rb_field` is defined.

## Changes

This PR fixes a performance issue with the calculation of the vector potential in the electromagnetic field calculation routine, moving the currently nested `rb_field` function out of the local scope of `em_field_calc` and into the module scope in `em_field_mod`.

Some notes:
* The cause of this slowdown is entirely due to how passing **internal procedures** as parameters to another procedure relies on creating a "trampoline."  Based on what I've read, the largest performance penalty then only applies to MacOS and more secure Linux configurations that don't allow an executable stack (see below).
* I'm not an expert with OpenMP (or any of this, to be fair). I know that RK tracking uses OpenMP to parallelize the calculations, so the promoted `rb_*` variables used by `rb_field` were marked thread-private with OMP. This could use some validation.

## Performance profiling notes

This is on MacOS, with a non-OpenMP build.

**Before**
Prior to this PR, loading a very large lattice with RK tracking:

image

**After**
After this PR
image

Tracking again happens after `bmad_parser2`, meaning the performance hit happens twice just in loading, leading to almost a penalty of 50% of the loading time. Our lattice parsing goes from 102 seconds down to 50 seconds.

### gcc background / OS-specific

Per `gcc` docs:

> `-ftrampoline-impl=[stack|heap]`
> 
>    By default, trampolines are generated on stack. However, certain platforms (such as the Apple M1) do not permit an executable stack. Compiling with -ftrampoline-impl=heap generate calls to __gcc_nested_func_ptr_created and __gcc_nested_func_ptr_deleted in order to allocate and deallocate trampoline space on the executable heap. These functions are implemented in libgcc, and will only be provided on specific targets: x86_64 Darwin, x86_64 and aarch64 Linux. PLEASE NOTE: Heap trampolines are not guaranteed to be correctly deallocated if you setjmp, instantiate nested functions, and then longjmp back to a state prior to having allocated those nested functions.

As in the note, because the trampoline can't be pushed to the stack, it instead goes on the heap, thrashing memory allocations during the `super_qromb` call.

### DIFF
--- a/bmad/code/em_field_calc.f90
+++ b/bmad/code/em_field_calc.f90
@@ -58,7 +58,8 @@ type (ele_pointer_struct), allocatable :: used_list(:)
 type (ele_struct), pointer :: lord, ele2
 type (ele_struct), optional :: original_ele
 type (lat_param_struct) param
-type (coord_struct) :: orbit, local_orb, lab_orb, lord_orb, this_orb
+type (coord_struct) :: orbit, lab_orb, lord_orb, this_orb
+type (coord_struct), target :: local_orb
 type (em_field_struct) :: field, field1, field2, lord_field, l1_field, mode_field
 type (cartesian_map_struct), pointer :: ct_map
 type (cartesian_map_term1_struct), pointer :: ct_term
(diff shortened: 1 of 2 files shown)


DavidSagan:
@ken-lauer Very interesting! Thanks for the PR and for diagnosing this.

PR #1905, opened 2026-01-29
Add an expression-checking test suite + fix a few bugs

ken-lauer:
## Background

I ran into a `modulo` bug (listed as 1 below) while loading up a lattice, and thanks to a debug Tao build the bounds checker showed where the cause was.

After finding and (I think) fixing it, I decided it was probably good to go through some basic expression tests in the suite. That uncovered a couple other issues.

(@DavidSagan - feel free to just accept the test suite and write your own fixes for the expression parser. I did what I thought seemed right based on the raised errors, but the tree evaluation engine logic is on the complicated side for me to fully grasp without a really deep dive.)

## Bugs/changes

1. `modulo` and `atan2`:

```
Tao> show value modulo(1, 2)
At line 413 of file /Users/klauer/Repos/bmad/tao/code/tao_evaluate_tree.f90
Fortran runtime error: Array bound mismatch for dimension 1 of array 'info_loc' (2/1)
```

This looks to be because the `info_loc` output isn't adjusted (starts off as 2 because of 2 inputs to `modulo`, but should be resized to 1 to represent the result - assuming I understand that correctly)

2. `ran`

Doesn't set the 'good' flag (unsure if this is important?)

Crashes with:

```
Tao> show value ran()
At line 175 of file /Users/klauer/Repos/bmad/tao/code/tao_evaluate_tree.f90
Fortran runtime error: Index '1' of dimension 1 of array 'info_loc' above upper bound of 0
```

Added a guard around this checking if 'ns > 0', which seems appropriate - but perhaps there's a better fix?

3. `ran_gauss`

Like `ran`, doesn't set the 'good' flag.

`ran_gauss` works with a parameter, but not without.

```
Tao> show value ran_gauss(1)
    -4.2623470997133039E-01

Tao> show value ran_gauss()
At line 373 of file /Users/klauer/Repos/bmad/tao/code/tao_evaluate_tree.f90
Fortran runtime error: Index '1' of dimension 1 of array 'tao_tree%node%value' above upper bound of 0
...
```

### DIFF
--- a/regression_tests/pytao/expressions/lat1.bmad
+++ b/regression_tests/pytao/expressions/lat1.bmad
@@ -1,0 +1,14 @@
+no_digested
+beginning[beta_a] = 10
+beginning[beta_b] = 10
+beginning[e_tot] = 10e6
+parameter[geometry] = open
+parameter[absolute_time_tracking] = T
+
+cav: Lcavity, voltage = 1e6, L = 1, rf_frequency = 1e9
+linac: line[multipass] = (cav)
+
+lat: line = (linac, linac)
+use, lat
+
+Lcavity::*[n_rf_steps] = 1024
--- a/regression_tests/pytao/expressions/test_expressions.py
+++ b/regression_tests/pytao/expressions/test_expressions.py
@@ -1,0 +1,187 @@
+from __future__ import annotations
+
+import pathlib
+from typing import cast
+
+import numpy as np
+import pytest
+from pytao import SubprocessTao, Tao
+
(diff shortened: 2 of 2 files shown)


DavidSagan:
@ken-lauer I put in a cleaner fix for this. What I don't understand is why my Tao regression tests did not fail. In any case, you should submit your tests in a PR.

ken-lauer:
Removed my fixes.

I believe since the test suites are running against production Bmad builds, they won't hit the bounds checker and all of the other safeguards that a debug build provides. The test suite was likely accessing invalid memory but just not outright failing.

PR #1935, opened 2026-02-20
Fix `ele` -> `half_ele` typo, causing intermittent particle loss

ken-lauer:
Looking at the surrounding code in `bmad/multiparticle/beam_utils.f90`, it's apparent that the radiation map should be set on `half_ele` rather than `ele`. This appears to have been a copy/paste mistake.

The change in `bmad/code/create_element_slice.f90` might not be strictly necessary, but for cleanliness it seemed correct to me.

@ChristopherMayes has an isolated test case that can trigger this bug.

### DIFF
--- a/bmad/code/create_element_slice.f90
+++ b/bmad/code/create_element_slice.f90
@@ -151,6 +151,10 @@ if (present(old_slice) .and. .not. rad_map_stale .and. .not. include_upstream_en
   case (sbend$, quadrupole$, sextupole$, octupole$, thick_multipole$)
     if (associated(sliced_ele%rad_map)) sliced_ele%rad_map%stale = .false.
   end select
+else
+  if (associated(sliced_ele%rad_map)) then
+    sliced_ele%rad_map%stale = .true.
+  endif
 endif
 
 ! For a sliced taylor element the %taylor%term components point to the lord components. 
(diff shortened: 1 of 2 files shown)


bmad/code/create_element_slice.f90
```diff
@@ -151,6 +151,10 @@ recursive subroutine create_element_slice (sliced_ele, ele_in, l_slice, offset,
   case (sbend$, quadrupole$, sextupole$, octupole$, thick_multipole$)
     if (associated(sliced_ele%rad_map)) sliced_ele%rad_map%stale = .false.
   end select
+else
+  if (associated(sliced_ele%rad_map)) then
+    sliced_ele%rad_map%stale = .true.
+  endif
```
DavidSagan: Slightly cleaner code:
```
if (associated(sliced_ele%rad_map) .and. present(old_slice) .and. .not. rad_map_stale .and. .not. include_upstream_end .and. &
    .not. include_downstream_end .and. ele0%value(l$) == sliced_ele%value(l$) .and. &
    (ele_in%tracking_method == bmad_standard$ .or. ele_in%field_calc == bmad_standard$)) then
  select case (ele_in%key)
  case (sbend$, quadrupole$, sextupole$, octupole$, thick_multipole$)
    sliced_ele%rad_map%stale = .false.
  case default
    sliced_ele%rad_map%stale = .true.
  end select
endif
```

DavidSagan:
@ken-lauer And thanks for the PR!

PR #1953, opened 2026-03-04
Fix for HDF5 search algorithm

jsberg-bnl:
Deals with two cases
1. FindHDF5 can't fun h5fc. FindHDF5 tries both spellings of the hl_fortran library, one of which is absent, and sets the imported target to the one not found. Showed up in a conda build with OpenMPI. Code fixes the bad hl_fortran imported target.
2. FindHDF5 in CMake before 3.19 doesn't even create imported targets. Showed up in an older RHEL8 install. Fixed by rolling our own target when there isn't one present.
I'm optimistically hopeful that this solves the HDF5 problems. But this is not the first time I've said that...


DavidSagan:
@jsberg-bnl Much thanks for the fix!

Issue #1963, opened 2026-03-11
`radiation_fluctuations` calcs spend >20% of compute time making random numbers


Nathan-Majernik:
When profiled with `gprof` it is observed that >20% of total compute time can be dedicated to generating random numbers. The attached minimal example gives instructions on building Bmad with `gprof` as well as lattice files for testing with and without radiation fluctuations.

A summary of results I produced from the profiling script included in the example is:

```
### Radiation ON (271s profiled time)

| % Time | Self (s) | Calls | Function | Category |
|--------|----------|-------|----------|----------|
| 11.7% | 31.7s | 2.6B | `ran_gauss_scalar` | **RNG** |
| 9.5% | 25.7s | 800M | `track1_radiation` | Radiation |
| 8.6% | 23.2s | 1.1B | `multipole_ele_to_ab` | Tracking |
| 6.4% | 17.4s | 1.1B | `max_nonzero` | Tracking |
| 4.2% | 11.4s | 5.5B | `pointer_to_ran_state` | **RNG** |
| 3.4% | 9.1s | 3.2B | `ran_uniform_scalar` | **RNG** |
| 2.2% | 5.8s | 1.6B | `ran_uniform_vector` | **RNG** |
| 0.9% | 2.4s | 0.4B | `ran_gauss_vector` | **RNG** |
| **Total RNG** | **60.4s** | | | **22.3% of runtime** |

### Radiation OFF (193s profiled time)

| % Time | Self (s) | Function |
|--------|----------|----------|
| 11.7% | 22.5s | `multipole_ele_to_ab` |
| 8.6% | 16.5s | `max_nonzero` |
| 5.8% | 11.3s | `track1_radiation` (damping only) |
| 3.9% | 7.6s | `check_aperture_limit` |
| — | ~0.1s | All `random_mod` functions combined |
```

[radiation_mwe.zip](https://github.com/user-attachments/files/25886376/radiation_mwe.zip)

DavidSagan:
Ziggurat now implemented.

PR #1964, opened 2026-03-11
Faster Gaussian RNG and add Ziggurat converter for radiation tracking

ChristopherMayes:
This addresses https://github.com/bmad-sim/bmad-ecosystem/issues/1963. 

### Summary

Optimizes `ran_gauss_scalar` and `ran_gauss_vector` in `sim_utils/math/random_mod.f90` to reduce the overhead of radiation fluctuation calculations, and adds a new `'ziggurat'` Gaussian converter option based on Marsaglia & Tsang (2000). On a 400-sbend lattice with 10k particles (10-run average on Apple Silicon):

- **Inlined Box-Muller** (default `exact` converter): reduces RNG overhead by **45%** (1.35s → 0.74s) with identical random sequences
- **Ziggurat converter**: reduces RNG overhead by **74%** (1.35s → 0.35s) — opt-in via `set global random_gauss_converter = ziggurat`

### Motivation

Profiling beam tracking with `bmad_com%radiation_fluctuations_on = T` showed that Gaussian random number generation (`ran_gauss`) was the dominant radiation-specific hotspot. The call chain `ran_gauss_vector` → `ran_gauss_scalar` (×6) → `ran_uniform` (×2 each) → `ran_uniform_scalar` → `pointer_to_ran_state` resulted in **12+ redundant function calls and pointer-to-state lookups per tracked element**.

Beyond call overhead, the Box-Muller rejection method requires `log()` and `sqrt()` on every accepted sample and has a ~21.5% rejection rate. The Ziggurat algorithm avoids these transcendental functions on ~97% of samples using a precomputed rectangular decomposition of the Gaussian PDF.

### Changes

**`sim_utils/math/random_mod.f90`**

**Inlined Box-Muller optimizations** (applied to default `exact_gaussian$` path):

**`ran_gauss_vector`** (the hot path — called from `radiation_mod` with a 6-element array):
- Resolve `pointer_to_ran_state` once instead of 12+ times per call
- Inline the Marsaglia xorshift + Park-Miller uniform RNG directly, eliminating calls through `ran_uniform_vector` → `ran_uniform_scalar`
- Batch Box-Muller pair generation to fill the array directly
- Copy `ix`, `iy`, `am` from the state struct to local variables before the loop for register promotion (the compiler cannot promote pointer-accessed struct members)
- When `sigma_cut` is active (rare), fall back to `ran_gauss_scalar` to preserve correct save/check behavior for rejected Box-Muller pairs
- Quasi-random and quick-Gaussian converters still fall through to the original per-scalar path

**`ran_gauss_scalar`**:
- Same inlined uniform RNG approach — eliminates 3 function calls (`ran_uniform` → `ran_uniform_vector` → 2× `ran_uniform_scalar` with `pointer_to_ran_state`) per Box-Muller iteration
- Same local-variable register promotion for state fields
- Quick-Gaussian and quasi-random paths are unchanged

**New Ziggurat converter** (`ziggurat$`):

- New `ziggurat$ = 5` constant and `'ziggurat'` name in `ran_gauss_converter_name`
- Module-level Ziggurat tables (`zig_xtab`, `zig_ytab`, 128 layers) initialized on first use via `zig_table_init()`
- Ziggurat paths in both `ran_gauss_vector` and `ran_gauss_scalar` with the same inlined Marsaglia+Park-Miller uniform RNG
- `ran_gauss_converter` subroutine updated to accept `'ziggurat'` as a valid converter name
- Uses the same underlying uniform RNG state — fully compatible with `ran_seed_put`/`ran_seed_get`, thread states, etc.
- Falls back to scalar path when `sigma_cut < 10` (same as Box-Muller)

### Behavioral guarantees

- **Inlined Box-Muller produces identical RNG sequences**: For the same seed, the exact same random numbers are produced (same algorithm, same constants, just inlined)
- **Ziggurat produces different sequences**: It uses the same uniform RNG but a different Gaussian conversion algorithm. Enabling it will change simulation results.
- **Quasi-random and quick-Gaussian paths are unchanged** — they still delegate to the original scalar-by-scalar code
- **Sigma-cut truncation** works identically for all converters
- **Ziggurat is opt-in** — the default converter remains `exact_gaussian$` (Box-Muller)

### How to enable Ziggurat

```
! In Tao:
set global random_gauss_converter = ziggurat

! In Fortran:
call ran_gauss_converter('ziggurat')
```

### Benchmark results

Test lattice: 400 sbends (l=0.5, angle=0.01) + drifts, 42 MeV electrons, 10k particles, single-threaded. Apple Silicon (M-series), gfortran 15.1 `-O2`.

#### End-to-end (Tao, 10-run average for Box-Muller, 5-run average for Ziggurat)

| Metric | Original | Inlined Box-Muller | Ziggurat | 
|---|---|---|---|
| Radiation ON | 4.03s ± 0.15s | 3.40s ± 0.13s | 2.85s ± 0.01s |
| Radiation OFF | 2.68s ± 0.02s | 2.65s ± 0.04s | 2.50s ± 0.01s |
| **RNG overhead** | **1.35s ± 0.15s** | **0.74s ± 0.14s** | **0.35s ± 0.01s** |
| **Overhead reduction** | — | **-45%** | **-74%** |

#### Standalone microbenchmark (4M vector calls, vec_size=6, `-O2`)

| Version | Time | Speedup vs Original |
|---|---|---|
| Original | 0.328s | — |
| Clean (structural only) | 0.312s | 1.05× |
| Inlined Box-Muller | 0.248s | 1.32× |
| **Ziggurat** | **0.129s** | **2.54×** |

A cleaner version using function calls (no inlining) was also benchmarked and achieved 29% reduction. The inlined version's additional gain comes from eliminating function-call overhead and enabling register promotion of RNG state variables — optimizations that gfortran `-O2` does not perform through pointer-accessed struct members. At `-O3`, the compiler inlines automatically and clean ≈ inlined.

### Regression test

Added `ran_gauss_test` to `regression_tests/`:
- **Deterministic reproducibility**: Fixed seed produces identical values for both scalar and vector paths
- **Scalar-vector equivalence**: Both paths produce the exact same sequence given the same seed
- **Statistical tests** (100k samples each for scalar and vector):
  - Mean ≈ 0 (z-test)
  - Variance ≈ 1
  - Skewness ≈ 0
  - Excess kurtosis ≈ 0
  - Anderson-Darling normality test (statistic well below 1% critical value)
- **Sigma-cut enforcement**: No samples exceed the truncation bound

> Note: The regression test validates the `exact_gaussian$` (Box-Muller) path, which remains the default. The Ziggurat converter produces statistically correct Gaussians (verified by the standalone benchmark's correctness checks) but with a different sequence.

### Files changed

| File | Change |
|---|---|
| `sim_utils/math/random_mod.f90` | Optimized `ran_gauss_scalar` and `ran_gauss_vector`; added Ziggurat converter (`ziggurat$`), tables, init routine, and paths in both scalar/vector |
| `regression_tests/ran_gauss_test/ran_gauss_test.f90` | New regression test (10 checks) |
| `regression_tests/ran_gauss_test/output.correct` | Reference output |
| `regression_tests/cmake_files/cmake.ran_gauss_test` | CMake build spec |
| `regression_tests/CMakeLists.txt` | Added `cmake.ran_gauss_test` |
| `regression_tests/TESTS.LIST` | Added `ran_gauss_test` |
| `regression_tests/test_fortran.py` | Added to pytest parametrize list |

> **Note**: `util/dist_prefs` has local build configuration changes (conda paths, plot package) that should **not** be included in the PR.

### Thanks

Thanks to Claude Opus 4.6 for this development!


regression_tests/ran_gauss_test/output.correct
ken-lauer, line 1: Was this written before "you" overhauled the algorithm or after?

DavidSagan: Looks like this is due to what I was doing at the same time. Since my changes and the changes here do not overlap, I suspect a merge to this PR from main will fix the issue.

DavidSagan: @ken-lauer @ChristopherMayes So I was wrong. The test failures are due to Claude.

DavidSagan: @ChristopherMayes Can Claude fix this?

ken-lauer: My original question was to see if the new changes are truly generating the same RNG sequence as the old code as indicated in the PR description. We'd need this file to be from the pre-PR version to be meaningful.

DavidSagan: If the new code is faster than the old code I think it worth using the new code even if the sequences are not the same. I will ask on SLACK if anyone cares.

sim_utils/math/random_mod.f90
```diff
@@ -246,16 +277,156 @@ subroutine ran_gauss_vector (harvest, ran_state, sigma_cut)
...
+      if (l_iy < 0) l_iy = l_iy + im_nr_ran
+      u1 = l_am * ior(iand(im_nr_ran, ieor(l_ix, l_iy)), 1)
+
+      l_ix = ieor(l_ix, ishft(l_ix, 13))
+      l_ix = ieor(l_ix, ishft(l_ix, -17))
+      l_ix = ieor(l_ix, ishft(l_ix, 5))
+      k = l_iy/iq
+      l_iy = ia*(l_iy - k*iq) - ir * k
+      if (l_iy < 0) l_iy = l_iy + im_nr_ran
+      u2 = l_am * ior(iand(im_nr_ran, ieor(l_ix, l_iy)), 1)
```
ken-lauer: There's 4 similar (maybe identical?) blocks here that are presumably unrolled "for speed" at the expense of readability. Two are here, the other two are in the "slow loop" section.

Actually, the loops themselves are identical I think?

Things like this you should typically leave up to the compiler to optimize rather than trying to outsmart it. It's likely worth a benchmark to see if a more maintainable version with function calls would actually be slower.

ChristopherMayes: Inlining like this seems to be the fastest method. I updated the PR summary with some experiments.

ChristopherMayes: See: https://gist.github.com/ChristopherMayes/5c158a7fa5e24665b205174b8a3e6b32

ChristopherMayes: @DavidSagan we could make it twice as fast this with the Ziggurat algorithm in the above gist. The caveat is that the random numbers will be different from previous Bmad versions.

ChristopherMayes:
@DavidSagan I added `ziggurat` in sim_utils/math/random_mod.f90:

1. New constant ziggurat$ = 5 and updated ran_gauss_converter_name array
2. Module-level Ziggurat tables (zig_xtab, zig_ytab) — 128-layer precomputed lookup, initialized on first use via zig_table_init()
3. Ziggurat path in ran_gauss_vector — branches before the Box-Muller path when gauss_converter == ziggurat$. Uses the same inlined Marsaglia+Park-Miller uniform RNG. Falls back to scalar for sigma_cut < 10.
4. Ziggurat path in ran_gauss_scalar — same algorithm, handles sigma_cut rejection
5. Updated ran_gauss_converter — accepts 'ziggurat' as a valid converter name

Users can enable it via:
```
Tao: set global random_gauss_converter = ziggurat
```
```
Fortran API: call ran_gauss_converter('ziggurat')
```

The existing exact_gaussian$ (Box-Muller) remains the default and is completely unaffected — regression test passes bit-identical.

ChristopherMayes:
I'm happy with the default being this in Bmad with `integer :: gauss_converter = ziggurat$`

regression_tests/ran_gauss_test/output.correct
ChristopherMayes: @ken-lauer yes, the refactor makes the same numbers.

ChristopherMayes: Ziggurat is different,

PR #1976, opened 2026-03-14
GPU-accelerated particle tracking via CUDA

Nathan-Majernik:
## Summary

This PR is an initial introduction of GPU-accelerated batch particle tracking to Bmad. When enabled, all particles in a bunch are tracked in parallel on an NVIDIA GPU through supported element types, with automatic fallback to CPU tracking for unsupported elements or configurations. The feature is fully opt-in at both build time and run time. More elements and conditions will be supported in future contributions.

Currently supported elements:
- **Drift** -- full phase-space update
- **Quadrupole** -- including misalignments, fringe fields, magnetic/electric multipole kicks, and low-energy z-correction
- **Sbend** -- including k1 map, general nonlinear bend map, multipoles with curvature corrections (not supported with `exact_multipoles`)
- **Lcavity** -- stair-step RF approximation for both standing-wave and traveling-wave cavities, including ponderomotive kicks (not supported with absolute time tracking, solenoid components, multipoles, coupler kicks, zero RF frequency, or zero length)

### Benchmark

The included stress test (`gpu_tracking_stress_test`) tracks 1M particles through a 17-element "kitchen sink" lattice that exercises all supported element types and parameter combinations: plain and apertured drifts, quadrupoles with misalignment/fringe/magnetic multipoles/electric multipoles and all-of-the-above, sbends with k1/fringe/misalignment/multipoles, and lcavities in standing-wave mode with phase offsets. Apertures are set to lose roughly 10% of particles. On an NVIDIA L4 GPU:

| | Time | Speedup |
|---|---|---|
| CPU | 29.4 s | -- |
| GPU | 4.6 s | **6.4x** |

GPU and CPU results agree to machine precision (max coordinate difference ~3e-15). All 1M particles have identical alive/lost state between the two paths.

## Design

GPU tracking operates at the bunch level inside `track1_bunch_hom`. The implementation follows a CPU-GPU sandwich pattern for quadrupoles and bends: misalignments and fringe fields are applied on the CPU, the element body is tracked on the GPU, then exit fringe and misalignment reversal are applied on the CPU. Both entrance and exit aperture checking is performed on the CPU around the GPU kernel.

Key components:
- `gpu_tracking_mod.f90` -- Fortran module with public API, SoA conversion, and dispatch logic
- `gpu_tracking_kernels.cu` -- CUDA kernels for supported element body tracking
- `beam_utils.f90` -- dispatch integration in `track1_bunch_hom`
- `bmad_struct.f90` -- `bmad_com%gpu_tracking_on` flag

The CUDA kernels cache device memory allocations across calls to avoid repeated allocation overhead. Particle data is converted from Bmad's array-of-structures (`coord_struct` array) to structure-of-arrays (flat `real(C_DOUBLE)` arrays) for efficient GPU memory access.

## How to use

**Build time:** Set `ACC_ENABLE_GPU_TRACKING=Y` in `util/dist_prefs`. Requires the CUDA Toolkit (`nvcc`, `libcudart`). The default is `N`. If the CUDA compiler is not found, the build proceeds gracefully without GPU support.

**Run time:** Either call `gpu_tracking_init()` (checks the `ACC_ENABLE_GPU_TRACKING` environment variable and probes for CUDA hardware), or set `bmad_com%gpu_tracking_on = .true.` directly. The `ele_gpu_eligible(ele)` function can be used to query whether a specific element supports GPU tracking.

When `gpu_tracking_on` is `.true.`, bunch tracking through eligible elements is automatically offloaded to the GPU. Unsupported elements, or bunches requiring radiation/spin/backward tracking, fall back silently to CPU tracking.

## Files changed

**Core implementation (new files):**
- `bmad/low_level/gpu_tracking_kernels.cu` -- CUDA kernels
- `bmad/low_level/gpu_tracking_mod.f90` -- Fortran module

**Integration into existing code:**
- `bmad/modules/bmad_struct.f90` -- added `gpu_tracking_on` to `bmad_common_struct`
- `bmad/multiparticle/beam_utils.f90` -- GPU dispatch in `track1_bunch_hom`, entrance/exit aperture checking
- `bmad/CMakeLists.txt` -- CUDA build support gated on `ACC_ENABLE_GPU_TRACKING`
- `util/dist_prefs` -- new `ACC_ENABLE_GPU_TRACKING` build preference

**Tests and benchmarks:**
- `regression_tests/gpu_tracking_test/` -- 26 test cases across all supported element types with various parameter combinations, plus stress test and bend benchmark (gated on `ACC_ENABLE_GPU_TRACKING` in CMake; commented out in `TESTS.LIST` since it requires GPU hardware)
- `util_programs/gpu_tracking_benchmark/` -- standalone benchmark program

**Documentation:**
- `bmad/doc/tracking.tex` -- new "GPU-Accelerated Tracking" section
- `bmad/doc/param-structs.tex` -- `gpu_tracking_on` parameter documentation
- `bmad/doc/subroutines.tex` -- `ele_gpu_eligible`, `gpu_tracking_init`, `gpu_tracking_reset` entries
- `bmad/doc/misc-programming.tex` -- GPU subsection under "Parallel Processing"
- `README.md` -- setup instructions

## Limitations

- Requires NVIDIA GPU and CUDA Toolkit
- Not presently compatible with radiation damping/fluctuations, spin tracking, or backward tracking (falls back to CPU)
- Lcavity does not support absolute time tracking, solenoid components, multipoles, coupler kicks, zero RF frequency, or zero length on GPU
- Sbend does not support `exact_multipoles` on GPU
- Aperture checking is performed at element entrance and exit only (not during body tracking)
- GPU tracking routines are not MP thread safe (only one CPU thread should call them at a time)


mattsignorelli:
Hi!

Are you aware of [SciBmad](https://github.com/bmad-sim/SciBmad.jl)? SciBmad is already fully GPU compatible, has been carefully optimized for zero branching inside kernels, and has many more features. To give you an idea, I can track 100,000 particles through a 9,000 element lattice, including time-dependent parameters ("rampers" in the classic Bmad sense), in **0.3 seconds per turn**. SciBmad also has batch _parameters_: you can make any lattice parameter (e.g. magnet strengths) be a vector of say a million different numbers, and a bunch of a million particles each sees that corresponding strength. It also is fully branchless in the parameters. I am currently doing brute force dynamic aperture optimizations for the EIC using this feature.

While I'm indifferent as to what gets merged here in classic Bmad, the GPU performance this PR achieves is still very poor compared to SciBmad, and so my recommendation would be to just use SciBmad instead. @DavidSagan has also just implemented his LCavity into SciBmad, so I think you should have everything you need.

PR #1977, opened 2026-03-15
Optimize FFT-based space charge calculation by caching

ChristopherMayes:
### Summary

Reduces wall-clock time of the FFT-based 3D space charge calculation by **~5x** at 10K particles and **~2.7x** at 100K particles on the `bmad-doc/tao_examples/space_charge` test case (32×32×64 mesh). Results are physically equivalent to the baseline.

### Changes

**`bmad/space_charge/fft_interface_mod.f90`**

- **Cache FFTW plans**: Previously, every call to `ccfft3d` created a new FFTW plan with `FFTW_ESTIMATE`, executed it, and destroyed it. With ~7 FFTs per field solve and ~3 field solves per adaptive step, this was ~21 plan create/destroy cycles per time step. Now, forward and backward plans are cached as `save` variables and only rebuilt when the mesh dimensions change (which almost never happens during a run).

- **Use `FFTW_MEASURE` instead of `FFTW_ESTIMATE`**: `FFTW_MEASURE` runs a brief benchmark on first plan creation to find the optimal FFT algorithm. This adds a one-time cost of ~0.5s but gives faster execution on every subsequent FFT.

- **One-time OpenMP thread setup**: `fftw_plan_with_nthreads` and `fftw_cleanup_threads` were being called on every `ccfft3d` invocation. Now the thread count is set once at initialization via `fft_cache_init()`.

**`bmad/space_charge/open_spacecharge_mod.f90`**

- **Persistent scratch arrays**: The `crho` and `cgrn` complex scratch arrays in `osc_freespace_solver2` (each 2N×2N×2N complex doubles, ~8 MB for the default mesh) were allocated and deallocated on every call. Now they are `save` variables that persist across calls and are only reallocated when the mesh size changes.

- **Lazy mesh resizing**: `deposit_particles` now keeps the mesh bounds stable across calls, only resizing when the bunch grows beyond the mesh or shrinks significantly (controlled by `mesh_growth_factor` and `mesh_shrink_factor`). When both factors are 0, the original tight-fit behavior is restored exactly. This keeps `delta` constant, enabling the Green function FFT cache below.

- **Green function FFT cache**: `osc_freespace_solver2` caches the forward-FFT of the Green function, keyed by `(delta, gamma, offset, mesh_size)`. The Green function depends only on grid geometry and relativistic gamma — not on the particle distribution — so it stays valid as long as the mesh doesn't resize. The charge density `rho` is freshly deposited from particle positions every step and FFT'd as usual. When the cache hits, the expensive Green function computation and its forward FFT are skipped entirely — only the rho FFT, pointwise multiply, and inverse FFT are needed.

- **Fused min/max computation**: `deposit_particles` now computes all 6 coordinate extrema (min/max of x, y, z) in a single pass over the particle arrays, replacing 6 separate `minval`/`maxval` calls.

- **OpenMP parallel deposition**: For large bunches (≥ 50K particles), `deposit_particles` uses thread-private rho arrays with a critical-section reduction. Below the threshold, the serial loop is used to avoid OpenMP overhead.

- **Batch field interpolation**: New `interpolate_field_batch` subroutine performs trilinear interpolation for an array of points with explicit scalar weights per field component (avoiding the `(:)` array-section overhead of the original). Includes `!$OMP PARALLEL DO` over the particle dimension. Bounds are clamped rather than checked with per-particle I/O.

- **Reduced crho zeroing**: `osc_freespace_solver2` now zeros only the padding region of the doubled complex array (the 7 octants outside `1:nx, 1:ny, 1:nz`), since the data octant is immediately overwritten by the rho copy. Saves 7/8 of a ~32 MB memset per call.

**`bmad/modules/bmad_struct.f90`**

- Added `mesh_growth_factor` (default 0.1) and `mesh_shrink_factor` (default 0.1) fields to `space_charge_common_struct`, so users can tune the lazy resize behavior via `set space_charge_com mesh_growth_factor = <value>`.

**`tao/code/tao_set_mod.f90`**

- Added `mesh_growth_factor` and `mesh_shrink_factor` to the Tao `set space_charge_com` handler.

**`tao/code/tao_show_this.f90`**

- Added `mesh_growth_factor` and `mesh_shrink_factor` to the `show global` and `show global -space_charge_com` output.

**`tao/code/tao_pipe_cmd.f90`**

- Added `mesh_growth_factor` and `mesh_shrink_factor` to the `pipe space_charge_com` output.

**`bmad/space_charge/space_charge_mod.f90`**

- `sc_field_calc` passes `space_charge_com%mesh_growth_factor` and `%mesh_shrink_factor` as optional arguments to `deposit_particles`.
- Field interpolation loop replaced with batch `interpolate_field_batch` call: coordinates are gathered, batch-interpolated (OpenMP parallelized), then results scattered back.

**`bmad/space_charge/csr_and_space_charge_mod.f90`**

- Both `deposit_particles` call sites in the `fft_3d$` path pass `space_charge_com%mesh_growth_factor` and `%mesh_shrink_factor`.
- **Extracted `apply_fft_3d_kicks`**: The `fft_3d$` space charge block (deposit, solve, interpolate, kick) is now a standalone subroutine, called from `csr_and_sc_apply_kicks`. Uses `interpolate_field_batch` for the interpolation phase, then applies kicks in a separate loop.
- **OpenMP parallel kick application**: The kick loop after field interpolation is parallelized with `!$OMP PARALLEL DO`. Each particle's kick is independent, so this scales with core count. Saves ~5s at 1M particles.

**`regression_tests/space_charge_test/space_charge_test.f90`**

- Sets `mesh_growth_factor = 0` and `mesh_shrink_factor = 0` at the start of the test to use the original tight-fit mesh behavior, ensuring bit-level reproducibility with the existing `output.correct`.

**Thread safety**: All module-level state mutations are guarded with `!$OMP CRITICAL` sections. Currently these routines are only called from serial code, but the guards protect against future use from parallel regions. FFTW plan creation/destruction and `fftw_plan_with_nthreads` are not thread-safe per FFTW documentation.

### Benchmark

Test: `cd bmad-doc/tao_examples/space_charge && tao -noplot -command "set beam_init n_particle = N;set space_charge_com ds_track_step = 0.001;quit"`
(32×32×64 mesh, 1m pipe, `fft_3d` space charge, adaptive time stepping)

| N particles | Before | After | Speedup |
|---|---|---|---|
| 10,000 | 31s | **6s** | **5.2x** |
| 100,000 | 41s | **11s** | **3.7x** |
| 1,000,000 | 155s | **64s** | **2.4x** |

`show beam end` output is physically equivalent before and after (emittance within ~0.04%).

# Perlmutter (NERSC) testing

[This is a similar benchmark](https://gist.github.com/ChristopherMayes/b694bec60363ccf22c38080a5e632a53) and relative speedup comparison for various numbers of threads on Perlmutter. 

bmad_pr_1977_perlmutter_speedup

### DIFF
--- a/bmad/modules/bmad_struct.f90
+++ b/bmad/modules/bmad_struct.f90
@@ -2169,6 +2169,10 @@ type space_charge_common_struct                   ! Common block for space charg
                                                   !   If a bin sigma is < cutoff * sigma_ave then ignore.
   real(rp) :: particle_sigma_cutoff = -1          ! 3D SC calc cutoff for particles with (x,y,z) position far from the center.
                                                   !  Negative or zero means ignore.
+  real(rp) :: mesh_growth_factor = 0.1             ! Fractional padding when growing SC mesh (default: 10%).
+                                                  !  Set to 0 for tight-fit (no caching speedup).
+  real(rp) :: mesh_shrink_factor = 0.1             ! Fractional threshold for shrinking SC mesh (default: 10%).
+                                                  !  Mesh shrinks when bunch fills < (1-this) of the mesh range.
   integer :: space_charge_mesh_size(3) = [32, 32, 64]  ! Gird size for fft_3d space charge calc.
   integer :: csr3d_mesh_size(3) = [32, 32, 64]         ! Gird size for CSR.
   integer :: n_bin = 0                            ! Number of bins used
--- a/bmad/space_charge/csr_and_space_charge_mod.f90
+++ b/bmad/space_charge/csr_and_space_charge_mod.f90
@@ -1466,7 +1466,8 @@ if (ele%space_charge_method == fft_3d$) then
     csr%position(n)%charge = p%charge
   enddo
 
-  call deposit_particles (csr%position(1:n)%r(1), csr%position(1:n)%r(2), csr%position(1:n)%r(3), csr%mesh3d, qa=csr%position(1:n)%charge)
+  call deposit_particles (csr%position(1:n)%r(1), csr%position(1:n)%r(2), csr%position(1:n)%r(3), csr%mesh3d, qa=csr%position(1:n)%charge, &
+    mesh_growth_factor=space_charge_com%mesh_growth_factor, mesh_shrink_factor=space_charge_com%mesh_shrink_factor)
   ! OLD ROUTINE: call space_charge_freespace(csr%mesh3d)
   call space_charge_3d(csr%mesh3d)
    
@@ -1728,7 +1729,8 @@ do i_step = 0, n_step
     csr%position(n)%r = p%vec(1:5:2)
     csr%position(n)%charge = p%charge
   enddo
-  call deposit_particles (csr%position(1:n)%r(1), csr%position(1:n)%r(2), csr%position(1:n)%r(3), csr%mesh3d, qa=csr%position(1:n)%charge)  
+  call deposit_particles (csr%position(1:n)%r(1), csr%position(1:n)%r(2), csr%position(1:n)%r(3), csr%mesh3d, qa=csr%position(1:n)%charge, &
+    mesh_growth_factor=space_charge_com%mesh_growth_factor, mesh_shrink_factor=space_charge_com%mesh_shrink_factor)  
 
   ! Give particles a kick
   ! TODO: simplify with fft_3d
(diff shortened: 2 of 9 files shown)


ChristopherMayes:
@DavidSagan I know I've complained about save statements in the past, but these seem useful, and they have `!$OMP CRITICAL` to guard against any multithreading collisions.

PR #1994, opened 2026-03-25
Claude fix to remove ld warnings on a Mac.

DavidSagan:
Documentation on fix:
[macos_deployment_target_fix.md](https://github.com/user-attachments/files/26254728/macos_deployment_target_fix.md)

This goes with PR <https://github.com/bmad-sim/bmad-external-packages/pull/12>

### DIFF
--- a/util/Master.cmake
+++ b/util/Master.cmake
@@ -333,6 +333,18 @@ ENDIF ()
 string (STRIP "${ACC_LINK_FLAGS}" ACC_LINK_FLAGS)
 
 IF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT $ENV{ACC_CONDA_BUILD} MATCHES "Y")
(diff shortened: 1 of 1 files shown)


jsberg-bnl:
I will give this a review on my system with MacPorts and HomeBrew.

However, my gut on this one is not to fix it. It looks like the linker basically does the right thing anyhow, so why not let it do so? This adds a bunch of code that is asking to break some day to make harmless compiler/linker warnings go away.

DavidSagan:
GCC V15 does not show any warnings so if your gut is still to not use this I am OK to close the PRs.

jsberg-bnl:
I think we should skip this, but I'll still run the tests and add a comment with the results in case we need to resurrect it.

jsberg-bnl:
Commenting for future reference, compile/run with the patch was successful on both MacPorts and HomeBrew.

PR #1998, opened 2026-03-26
Let Tao use `transfer_fieldmap`

### LINKED ISSUE #1997: Feature request: Tao command to share fieldmap data between elements
## Problem

Bmad's parser loads a separate copy of HDF5 fieldmap data for each element with `grid_field = call::file.h5`, even when multiple elements reference the same file. For lattices with many identical maps, this causes memory to scale as O(n_elements) rather than O(n_unique_files)

## Existing infrastructure

Bmad already has `transfer_fieldmap` (`bmad/low_level/transfer_fieldmap.f90`) which does reference-counted pointer sharing of fieldmap data between elements via `%ptr` and `%n_link`. This works correctly for multipass lord/slave relationships but is not accessible from Tao's command interface.

## Request

Expose `transfer_fieldmap` through a Tao command, e.g.:

```
set element <ele_out> fieldmap = <ele_in>
```

This would allow users to:
1. Load a fieldmap on one element per unique HDF5 file
2. Share the data to all other elements of the same type via the Tao command line or pytao

Nathan-Majernik:
## Add `set element <ele> fieldmap = <source>` Tao command

Exposes Bmad's existing `transfer_fieldmap` through the Tao command interface. This lets users load a fieldmap once and share it to other elements via pointer sharing, rather than loading redundant copies from disk.

### Changes

- **`tao/code/tao_set_mod.f90`** -- Intercept `attribute == "FIELDMAP"` in `tao_set_elements_cmd`. Locates the source element, calls `transfer_fieldmap(..., all$)` for each target, then marks the lattice for recalculation.
- **`tao/doc/command-list.tex`** -- Document the new `fieldmap` pseudo-attribute in the `set element` section.
- **`bmad-doc/lattices/transfer_fieldmap_example/`** -- Minimal example: three solenoids, one with an inline grid_field, two receiving it via the new command.

### Usage

```
set element sol_b fieldmap = sol_a          ! single element
set element solenoid::* fieldmap = sol_a    ! wildcard
```

Closes #1997

### DIFF
--- a/bmad-doc/lattices/transfer_fieldmap_example/lat.bmad
+++ b/bmad-doc/lattices/transfer_fieldmap_example/lat.bmad
@@ -1,0 +1,44 @@
+! Minimal lattice demonstrating fieldmap transfer between elements.
+! sol_a has an inline grid_field; sol_b and sol_c are identical solenoids
+! whose fieldmap data will be shared from sol_a via the Tao command:
+!   set element sol_b fieldmap = sol_a
+!   set element sol_c fieldmap = sol_a
+
+no_digested
+
+parameter[geometry] = open
+parameter[p0c]      = 1e9
+parameter[particle] = electron
+
+beginning[beta_a] = 10
+beginning[beta_b] = 10
+
+d1: drift, l = 0.5
+
+sol_a: solenoid, l = 1,
+  tracking_method   = runge_kutta,
+  mat6_calc_method  = tracking,
+  field_calc        = fieldmap,
+  grid_field = {
+    geometry   = rotationally_symmetric_rz,
+    field_type = magnetic,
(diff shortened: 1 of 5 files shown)


DavidSagan:
Why not just automatically share the field map?

Nathan-Majernik:
> Why not just automatically share the field map?

How exactly? I thought Bmad's `transfer_fieldmap` was the mechanism for sharing?

DavidSagan:
By default there should be sharing. Not sharing looks like a bug to be fixed. Please open an issue with a simple example.

Nathan-Majernik:
> By default there should be sharing. Not sharing looks like a bug to be fixed. Please open an issue with a simple example.

Done #1999

PR #2013, opened 2026-04-06
Remove incorrect CMAKE_*IGNORE_PATH settings

jsberg-bnl:
**Do not merge this yet.** I need to test this across platforms first, though I'm almost certain it will be fine. 

These are wrong, so I'm removing them, but someone put this hack in to solve a problem, please check those tickets in case we need to address that problem before merging. Might need to check with SLAC also.

### DIFF
--- a/util/Master.cmake
+++ b/util/Master.cmake
@@ -73,14 +73,6 @@ ELSE ()
   set (PACKAGES_DIR ${RELEASE_DIR})
 ENDIF ()
 
-# Explicitly remove 32-bit Libraries from Linux build PATH for 64-bit builds - RT#43178
-# Added /lib to further restict 32-bit Linux build PATH - RT#56203
-IF (${CMAKE_SYSTEM_NAME} MATCHES "Linux" AND NOT "$ENV{ACC_FORCE_32_BIT}" MATCHES "Y")
-  SET (CMAKE_IGNORE_PATH /lib /usr/lib)
-  SET (CMAKE_SYSTEM_IGNORE_PATH /lib /usr/lib)
-ENDIF ()
-
-  
 IF (FORTRAN_COMPILER MATCHES "gfortran")
 
   SET (RELEASE_NAME $ENV{DIST_BASE_DIR})


DavidSagan:
Ticket original problem:
```
I am attempting to install bmad on the fermilab cluster so that our g-2 collaborators can use the simulation code that we have written. I get the following error. The file /nashome/d/drubin/bmad_dist/production/gnu_utilities/log/texinfo-4.8.log is also attached.

Suggestions?
Thanks
David Rubin

-bash-4.1$ ./util/dist_build_production 
Building directory: plplot

Preparing production build of PLplot VERSION 5.9.9 Oct 12 2011...

Using: gcc (GCC) 4.9.3
Using: GNU Fortran (GCC) 4.9.3
Checking minimium version of GNU Utility Programs needed...

Using found libtool-2.4.2
Using found m4-1.4.16
Using found automake-1.12.2

Cannot check version of makeinfo 
The installed version of makeinfo maybe too old.

The installed version of makeinfo cannot be used.
Attempting to build texinfo-4.8...
Error in building texinfo-4.8, Please see /nashome/d/drubin/bmad_dist/production/gnu_utilities/log/texinfo-4.8.log. 
Error in building PLplot VERSION 5.9.9 Oct 12 2011.

Error in distribution build.
-bash-4.1$ Write failed: Broken pipe
```

jsberg-bnl:
@DavidSagan, are you sure that's the right one? I can't see how that would be connected to this bit of code. From the email you forwarded, this appears to be 67716, rather than 43178 and 56203 that this refers to.

DavidSagan:
The email I forwarded is not related. I only pasted the beginning of the correspondence for 43178. I finally figured out how to extract the whole email chain:
[43178.txt](https://github.com/user-attachments/files/26544418/43178.txt)

jsberg-bnl:
Good, thanks. This is sufficiently vintage that I think we can ignore it. Part of what is going on might be a self-built cmake which doesn't have awareness of the system library layout.

DavidSagan:
So should I merge now?

jsberg-bnl:
Not yet, I want to test this on several other distributions, I'm hoping to get to it tomorrow. So far I'm the only one affected by this, and the conditions for someone else to trip it are pretty narrow (cutting-edge distribution like Arch, Manjaro, Fedora, Debian, though there's a danger of this showing up in Macports/Homebrew soon, and using my system for building against system packages).

jsberg-bnl:
And it can't trip on a Mac.

jsberg-bnl:
This succeeded on all systems I tested, so please merge.

PR #2020, opened 2026-04-14
Allow symbolic number names in datum `meas` fields

### LINKED ISSUE #2019: Feature request: Allow symbolic number names in the `meas` field of datum definitions in tao.init namelists.
Currently `meas` only accepts literal reals, so the only way to use named targets is the `expression:` workaround in `data_type`, which loses the semantic data type (e.g. beta.a) and requires a more verbose per-datum syntax. Resolving symbolic names in `meas` would allow the clean form:  

```
&symbolic_number TARGET_END_BETA_A = 1.0 /                                                                                            
          
...
                                                                                                                            
datum(1) = 'beta.a' '' '' 'END' 'target' TARGET_END_BETA_A 1e1                   
```                                              
                                                                                                                                      
The &symbolic_number preprocessor already runs before namelist reading, so the values are in memory, they just need to be substituted into (or resolved after) the namelist parse of `meas` fields.

Nathan-Majernik:
Pure Claude code... but it seems like it works? Closes #2019

## Summary
Preprocesses data init files to substitute &symbolic_number names with their numeric values before Fortran namelist reads

Enables, e.g., 
```
&symbolic_number TARGET_END_BETA_A = 1.0 /                                                                                            
          
...
                                                                                                                            
datum(1) = 'beta.a' '' '' 'END' 'target' TARGET_END_BETA_A 1e1                   
```         

Substitution is now skipped when the matched name is:
  – Preceded by % (component access, e.g. datum%meas)
  – Followed by = (namelist variable assignment, e.g. default_meas = ...)
  – Followed by ( (array/function name, e.g. datum(...))
This prevents collisions with namelist keywords, struct components, and array names.


DavidSagan:
I'm leery of this since it only solves one small part of the problem of the lack of expression evaluation in namelists. A more general solution of using something like a python preprocessor I think would be better.

Nathan-Majernik:
> I'm leery of this since it only solves one small part of the problem of the lack of expression evaluation in namelists. A more general solution of using something like a python preprocessor I think would be better.

How about this version?

PR #2032, opened 2026-04-30
Added `-version` command line parameter to Tao.

DavidSagan:


DavidSagan:
@ChristopherMayes @ken-lauer PyTao tests are failing since I change the version string output from `Date: <version>` to `Version Date: <version>` to make things clearer. I can change back if you want or you can update PyTao. What do you want?

ken-lauer:
It's easier for my lazy self to not have to update the mechanism in PyTao, of course!

If you really want the output to be changed I can address it, of course. It would take a bit for that fix to propagate, so the test would continue to fail until that point.

It's entirely up to you - just let me know.

DavidSagan:
No biggie. I will revert.

PR #2041, opened 2026-05-11
Fix inum ix_branch off-by-1 error

ken-lauer:
* `branch` is `branch_struct(0:)` (i.e., 0 indexed)
* Currently: in the single branch case, `0` to `size(branch)` gives `[0, 1]` erroneously - it should be just `[0]`
* Fix: `lbound` to `ubound` is guaranteed to give the right indices by iterating explicitly through the bounds
   * Technically could have left this to be `0` to `ubound`, but this seemed cleaner


ken-lauer:
Test suite errors appear unrelated to this PR, failing with some very slightly differing emittance measurement changes just beyond the configured tolerance.

DavidSagan:
> Test suite errors appear unrelated to this PR, failing with some very slightly differing emittance measurement changes just beyond the configured tolerance.

I agree. But why would these two tests fail and not the others?

DavidSagan:
One possibility is to increase the tolerance on the test.

ken-lauer:
> I agree. But why would these two tests fail and not the others?

My guess is that some updates to the build toolchain on the conda jobs have resulted in slightly different test suite outputs. Checked in #2042 for good measure that this PR's changes didn't cause this (regardless of the minimal likelihood).

libopenblas, libgcc, and gfortran all had minor updates between the last passing master build and now. The tolerance is pretty close to the difference:

```
   Data from "output.now":     [3.71196238e-07, 2.66477176e-08, 1.68959115e-06]
   Data from "output.correct": [3.71196247e-07, 2.66477181e-08, 1.68959118e-06]
   Diff: 3.000000010882481e-14 Diff/Val: 1.7755774728394023e-08
```

with
```
"init-emit" REL 1E-8  3.71196247E-07  2.66477181E-08  1.68959118E-06
```

Perhaps it was always just on the edge of being acceptable?

DavidSagan:
> Perhaps it was always just on the edge of being acceptable?

I checked the regression results with ifort and gfortran (V 13.3.1). The values are not on the edge. Tracking this down would be interesting but could take a lot of time (which I for one do not have). The easy fix would be to increase the test tolerance. What do you think?

ken-lauer:
Behind the scenes today, I tried to pin it down to a specific dependency change that started to cause the difference in resulting values. The ones you'd expect to have some sort of change like libblas or gfortran version bumps were all in a sizable GitHub Actions matrix for a build. The results overall were not always consistent; one time a held-back `libblas` passed and the next time it didn't.

I also rolled back to roughly the last 30 commits to see if any recent changes caused the failure in combination with the new build environment, but failures were somehow consistent there. In other words, no specific recent commit caused this.

Floating point values generated by way of many different mathematical operations require some real analysis to see what tolerances are actually appropriate. 

In any case, I'm afraid my skills in that regard are limited, so I took your suggestion and bumped the tolerance to 1e-6 and we're back to passing tests.

DavidSagan:
Can you change this to 1e-7 instead of 1e-6?

PR #2064, opened 2026-06-23
allow alias to be set for overlay elements

MichaelEhrlichman:
TYPE, ALIAS, and DESCRIP are listed in the Bmad manual as valid parameters for groups and overlays.  Also, they can be set OK via the lattice file.  However, tao says they are not free.

The not free error appears to be a bug, which this PR fixes.


DavidSagan:
@MichaelEhrlichman This was only part of the problem. #2066 fixes everything. So I will close this in favor of #2066

DavidSagan:
@MichaelEhrlichman But thanks for trying.

Issue #63, opened 2020-05-12
Extended CSR calculation for a wiggler


DavidSagan:
Problem: Right now, the calculation of CSR in a wiggler can only be done by having the user split the wiggler into slices so that the spline fit is accurate. 

Solution: To alleviate the burden on the user, the CSR calculation should do the splitting. This can be done by modifying the csr_ele_info_struct to contain an array of substructures. With this, particle position is indexed by element index and index in substructure array.

Also: To save on computation time, instead of passing just the centroid from track_beam to track1_bunch_csr, the csr variable could be passed with %eleinfo precomputed. It is not clear how much time this would save so profiling should be done first.

Also: It is not clear that the computation of the centroid spline needs to extend out to n_ele_track rather than the current element (or perhaps a few elements afterwards).

DavidSagan:
PR #2085 covers this.

PR #2085, opened 2026-07-04
Auto-subdivide wigglers in the 1D CSR calculation (issue #63)

DavidSagan:
Addresses #63 (the wiggler-subdivision part; the two "Also" optimizations are left for follow-up).

## Problem

The 1D CSR calculation in `csr_and_space_charge_mod.f90` represents each lattice element by a single centroid **chord** (a straight line from the centroid at the element entrance to the centroid at the exit) plus a spline giving the centroid's transverse deviation from that chord. For a bend or drift this is accurate. For a **wiggler/undulator** the centroid orbit oscillates *within* the element, so one entrance-to-exit chord+spline is a poor fit — today the user must hand-split the wiggler into many short elements, and the code emits a warning that the spline fit is inaccurate.

## Solution

The CSR calculation now does the splitting itself.

- New `csr_chord_struct` holds the per-chord geometry (`spline`, chord-end floor positions, `theta_chord`, `L_chord`, `dL_s`, element-ref-s span). `csr_ele_info_struct` keeps the element-level data and gains a `sub(:)` array of chord segments.
- A source/kick position is now addressed by `(element index, chord-segment index)`. The index is threaded through `csr_kick1_struct` (`ix_sub_source`) and `csr_struct` (`ix_sub_kick`).
- **Wigglers/undulators** (planar or helical model, `csr_method = one_dim`) are subdivided at the CSR step size. The interior sub-segment boundary centroid positions are obtained by **internally tracking a single centroid particle** through the element (`element_slice_iterator` + `track1`). Element-boundary geometry is kept exactly equal to the passed-in `centroid` array, so only *interior* resolution is added.
- The routines that walked element-by-element (`s_source_calc`, `ddz_calc_csr`, `I_csr`, `image_charge_kick_calc`, the kick-point setup, `s_ref_to_s_chord`) now walk chord segments, via small `prev_seg`/`next_seg`/`Ls_start_to_kick` helpers.
- The obsolete "spline fit is inaccurate for a wiggling orbit" warning is removed.

## Key invariant

Every element except a subdivided wiggler gets **exactly one** chord segment. With one segment the `(ele, sub)` walk reduces algebraically to the previous element walk, so **all non-wiggler results are unchanged** — this is the safety net for the whole refactor.

## Verification

- `bmad` builds and links clean.
- The `csr_and_space_charge` regression is **bit-identical** to the previous reference for the existing (non-wiggler) cases.
- An added planar-model-wiggler case subdivides correctly (L = 0.5 m, ds = 0.01 m → 50 chord segments) and tracks 1000/1000 particles without error.

## Please note

I can't independently validate the wiggler CSR *numbers* here — there's no known-good wiggler-CSR benchmark in the suite. The new `Wiggler-CSR` reference line is a **behavior-lock baseline** (it catches unintended changes to the subdivision machinery); it is **not** a physics validation. That check is for a domain expert to make.

## Design choices (as agreed)

- Intra-element centroid comes from **internal tracking** (no API change to callers).
- **Nested sub-array** layout in `csr_ele_info_struct` (as the issue describes).
- **Subdivision only** — the issue's two "Also" items (precompute `%eleinfo` in `track_beam`; trim the `n_ele_track` look-ahead, which merits profiling first) are left for follow-up.


ChristopherMayes:
# Review of PR #2085 — Auto-subdivide wigglers in the 1D CSR calculation

## Overview

The PR moves per-chord geometry out of `csr_ele_info_struct` into a new `csr_chord_struct` with a
`sub(:)` array, subdivides planar/helical-model wigglers at the CSR step size using an internally
tracked centroid particle, and converts all the element-walking routines to walk
`(element, segment)` pairs.

The refactor itself is solid: the single-segment path was verified to reduce exactly to the old
code (including the subtle ordering in `set_chord`, the `i == 0` self-reference, and error
propagation), no symbol or struct changed by the PR is used outside the module, and the
`runt`/`track1` aliasing patterns match established bmad idioms.

Review method: 8 independent finder angles (line-by-line diff scan, removed-behavior audit,
cross-file trace, reuse, simplification, efficiency, altitude, conventions), candidates deduped,
then each surviving candidate independently verified against the code.

## Findings (most severe first)

### 1. Pre-existing: linear energy interpolation has swapped weights

`bmad/space_charge/csr_and_space_charge_mod.f90:356`

`f1 = s0_step / ele%value(l$)` is 0 at the element entrance, but

```fortran
e_tot = f1 * branch%ele(ele%ix_ele-1)%value(e_tot$) + (1 - f1) * ele%value(e_tot$)
```

weights the *entrance* energy by `f1` — so at the entrance the exit energy is used and vice versa.
For CSR through an energy-changing element (lcavity, e_gun), `csr%gamma2` is computed with the
energy profile mirrored, giving wrong `I_csr` and shielding everywhere the energy changes.
Untouched by this PR, but it sits in the middle of the touched step loop and is a one-line fix.

### 2. Subdivision is gated on the wiggler's own `csr_method`, but upstream wigglers act as source geometry regardless

`bmad/space_charge/csr_and_space_charge_mod.f90:272`

A user who enables 1D CSR only on a bend/chicane downstream of a planar-model wiggler (natural —
they don't want CSR kicks *inside* the wiggler) gets the wiggler represented by the old single
entrance-to-exit chord when `s_source_calc` walks source points back into it — exactly the
issue-#63 inaccuracy, silently. This is a carried-forward gap (the old warning also only fired for
the tracked element), but the fix machinery now exists and dropping the `csr_method` test from the
gate would close it.

### 3. Subdivision can silently not engage, and the old warning was removed unconditionally

`bmad/space_charge/csr_and_space_charge_mod.f90:275`

Three paths give `n_sub = 1` with no diagnostic:

- **(a)** `ds_sub` resolves to 0 for a *source* wiggler (its `csr_ds_step` unset and
  `space_charge_com%ds_track_step` defaults to 0 at `bmad_struct.f90:2162` — the fatal check at
  line 230 only validates the tracked element);
- **(b)** a coarse step, e.g. `csr_ds_step = 0.6` on a 0.5 m wiggler;
- **(c)** more generally, granularity is keyed to the CSR kick step, not the wiggler period — a
  segment spanning several periods has a spline as bad as the old per-element one
  (`l_period$ = l/n_period$` is available to check).

Suggest warning when a planar/helical wiggler ends up with segments longer than ~a period.

### 4. The centroid-sampling `track1` call omits `ignore_radiation = .true.`

`bmad/space_charge/csr_and_space_charge_mod.f90:300`

`track1`'s header says this flag exists precisely for reference-particle tracking. With
`bmad_com%radiation_fluctuations_on` (routine for machines that have wigglers), the sampling
particle receives stochastic kicks, making the interior chord geometry noisy run-to-run and
perturbing the global RNG stream; damping biases interior points while the endpoints stay pinned to
the passed-in centroid, creating an artificial kink in the last segment.

### 5. `orbit%state` is never checked in the subdivision loop

`bmad/space_charge/csr_and_space_charge_mod.f90:300`

`track1` silently returns the frozen orbit for a dead particle (verified: dead-orbit early return
with `err_flag = .false.`), so if the centroid clips an aperture mid-wiggler — possible even when
whole-element tracking of the passed-in centroid succeeded, since the oscillating orbit is now
checked at every slice boundary — all remaining sub-chords collapse toward the loss point:
silently corrupted CSR geometry or an opaque "VERY CONFUSED CSR" fatal with no hint of the cause.

### 6. Pre-existing: roundoff-recovery paths assign `s_source` the *value* of `ddz_calc_csr` instead of the root position

`bmad/space_charge/csr_and_space_charge_mod.f90:983`

In the `last_step == -1` branch the root is at the segment exit, so `s_source` should be
`esub%L_chord` (the `ddz_calc_csr` call is still needed for its side effects on `kick1`); instead
`s_chord_source` ends up ≈ 0 (a length mismatch of roundoff size), wrong by a full segment.
Downstream, `image_charge_kick_calc` evaluates the spline curvature at the wrong end — inside a
subdivided wiggler the curvature there can differ in sign, and subdivision multiplies the segment
boundaries where this path fires. (The mirror branch at line 988 is accidentally near-correct
since the true position is 0.)

### 7. For helical-model (and tilted planar) wigglers, subdivision only captures the horizontal projection of the wiggle

`bmad/space_charge/csr_and_space_charge_mod.f90:2006`

All chord geometry is forced into the y = 0 plane (`fl%r(2) = 0`; `L_chord`/splines use
`r(1)`,`r(3)` only), so the vertical oscillation's path-length excess is still missing — roughly
half the effect for a helical device, all of it for a `tilt = pi/2` planar one. This is inherent to
the 1D projected model (pre-existing), but the PR enables subdivision for `helical_model` and
removes the "may be invalid" warning, implying the problem is fixed. One for the domain-expert
physics pass the PR notes already request.

### 8. Dead local declarations left by the refactor

`bmad/space_charge/csr_and_space_charge_mod.f90:174`

In `track1_bunch_csr`: `vec0`, `vec`, `theta0`, `theta1`, `L`, `floor`, `parallel0`, `parallel1`,
and `ie` are now unused (the only `vec`/`floor` hits in the body are `c0%vec(6)`,
`branch%ele(i)%floor`, and comments); `i` in `s_source_calc` is also declaration-only now. Trivial
deletion.

## Weak / optional candidates (not in the ranked list)

- The step-size fallback pattern (`csr_ds_step$` → `space_charge_com%ds_track_step` →
  `max(1, nint(L/ds))`) now appears in three places in the module. A tiny shared helper is
  possible, but the new site's silent-vs-fatal difference from the other two is intentional.
- The `n_sub == 1` branch could be merged into the general loop with an `if (j < n_sub)` tracking
  guard — verified bit-identical — but the explicit branch is arguably load-bearing documentation
  for the PR's "non-wiggler results unchanged" safety claim.
- Efficiency: the `Ls_start_to_kick` sum in `s_source_calc`'s lattice-start branch is a per-step
  constant recomputed per kick bin and would be an easy cache. Deeper optimization of the O(n_sub)
  segment walks (`ddz_calc_csr`'s cross-segment loop is the `super_zbrent` objective) belongs with
  the profiling follow-up the PR already defers.

PR #2093, opened 2026-07-07
Performance tweaks for attribute bookkeeping

ken-lauer:
## Background

I'm trying to update thousands of attributes across many elements programmatically and run the lattice bookkeeper.
Profiling a reasonably simple benchmark shows that:
* A very significant period of time is spent on string comparisons (`select case`, `attribute_index`)
* `attribute_free` calls end up taking a lot of time, even if I know ahead of time that the attribute is free
* Overlays that are off still run through a lot of attribute index checks unnecessarily (including a ton of string comparisons)

### Changes

This PR mainly addresses the string comparison bits.
* Instead of comparing attribute names, it changes over to attribute constants.
* `attribute_index` is avoided in a few hot spots
* It adds an `ix_attrib` input for `attribute_free` to avoid yet another string lookup of the attribute
* It avoids a single case of `pointer_to_attribute` when a "regular" attribute is found as it's a rather simple operation to update the all_pointer_struct
* It avoids further bookkeeping when an overlay is found to be off ([ee3ed21](https://github.com/bmad-sim/bmad-ecosystem/pull/2093/commits/ee3ed219480d3bcd157190b59dd9509f1c515297))

### DIFF
--- a/bmad/code/attribute_bookkeeper.f90
+++ b/bmad/code/attribute_bookkeeper.f90
@@ -57,6 +57,7 @@ type (converter_prob_pc_r_struct), pointer :: ppcr
 type (molecular_component_struct), allocatable :: component(:)
 type (material_struct), pointer :: material
 type (material_struct) :: materi
+type (ele_attribute_struct) ds_step_info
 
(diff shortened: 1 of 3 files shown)


ken-lauer:
@DavidSagan when you get a chance to take a look at this, I would appreciate it.

PR #2096, opened 2026-07-09
Fix PTC radiation-map crash (DA VECTOR -9) from the #2094 forest resync

DavidSagan:
## Summary

The forest resync merged in #2094 ("Implemented PTC Pancake tracking with new Generalized Gradient code") introduced a memory-management regression in PTC that crashes ordinary radiation-map setup, plus two regression-test baselines that needed updating (one a genuine physics correction, two floating-point relocations). This PR fixes the crash and refreshes the baselines. `ptc_test`, `mat6_calc_method_test`, and `long_term_tracking_test` all pass again.

## The crash (main fix)

`radiate_2_probep` in `forest/code/Sh_def_kind.f90` received a mismatched alloc/kill refactor:

| local | `alloc` scope | `kill` scope (as merged in #2094) |
|-------|---------------|-----------------------------------|
| `z`   | inside the `kind22` branch only | **unconditional** (routine tail) |
| `av`  | unconditional | inside the `kind22` branch only |

Previously `alloc(z)` and `kill(z)` were **both** inside the `kind22` branch (balanced). The refactor moved `kill(z)` out to the unconditional tail without moving `alloc(z)`, and left `kill(av,3)` stranded inside `kind22` even though `alloc(av,3)` had become unconditional.

### Failure chain

For any **non-`kind22`** element (quadrupole, sbend, drift, rfcavity — essentially every ordinary lattice), `call kill(z)` runs on an **unallocated** `REAL_8`, freeing a stale DA handle:

1. The double-free makes `dadal1` do `nhole = nhole + 1` without creating a real hole, so the free-slot count desyncs from `allvec`.
2. A later `daall1` trusts `nhole > 0` and searches `allvec` **downward** for a free slot that doesn't exist; the loop has no lower bound, so `ind` underflows past 1 to **`-9`**.
3. That yields an out-of-bounds write `allvec(-9) = .true.`, then `daclr(-9)` then `dainf(-9)`, which prints `ERROR IN DAINF, DA VECTOR -9 NOT FOUND` and blocks in `mypauses` reading stdin (the apparent "hang/pause").

The bug is **release-only**: at `-O0` the stale handle is benign, and it surfaces far from its cause — in the `xpmap**(-1)` map inversion inside `ptc_setup_map_with_radiation` (`map_order = 1`, radiation + spin). It manifested as `ptc_test` aborting/pausing on macOS production builds.

### Fix

Swap the two kills so each matches its `alloc` scope: `kill(z)` back inside `kind22`, `kill(av,3)` at the unconditional tail. This also removes the `av` leak on the non-`kind22` path. The change is pure memory management and does not alter any tracking result.

## Secondary latent fix

`forest/include/a_def_all_kind.inc`: the `ekick` pointer added to `CAV4`/`CAV4P` in #2094 lacked `=> null()` — the only pointer in those types without default initialization (even the `symplectic` pointer added in the same commit has it), leaving its association status undefined. Added the initializer. This was not the `ptc_test` crash cause but is a real defect.

## Regression baseline updates

- **`mat6_calc_method_test`** — a genuine **correction**. `EM_FIELD1`/`EM_FIELD2` are generalized-gradient elements tracked with `symp_lie_ptc` (the PTC pancake path). The old baseline held pre-#2094 non-symplectic matrices (e.g. the (5,6) / T566 term ~`0.714` vs `runge_kutta` `0.361`, symplectic error ~`9e-5`). With the new pancake code these now agree with `runge_kutta` to ~`1e-8` and are symplectic to ~`2e-12`. Only `EM_FIELD1/2` changed above tolerance; all other elements are sub-tolerance.
- **`long_term_tracking_test`** — the `sim2` (PTC-tracking) block shifted at the ~`1e-8` level from the resync's floating-point reorganization; the `sim1` (BMAD-tracking) block is byte-identical.
- **`ptc_test`** — the `layout-sigma1` stochastic-sigma datum shifted ~`3e-12` from the same resync; regenerated so the now-non-crashing test passes.

## Testing

Built forest at `-O2` (production) and ran the affected tests: `ptc_test` (0 failures), `mat6_calc_method_test`, and `long_term_tracking_test` all pass.


DavidSagan:
@ken-lauer @ChristopherMayes PyTao test still failing. Is there something you need to do on your side?

ken-lauer:
Yes.

1. Waiting for bmad release to go through conda-forge
2. Then I will merge the PyTao fix for the new bmad version to `main`

DavidSagan:
@ken-lauer This PR fixes some regression test failures. Should I merge now or wait?

ken-lauer:
May as well merge, as it doesn't change the state of the already-failing PyTao test suite.

PR #2097, opened 2026-07-09
Fix bug using Fable as requested by Dave

ChristopherMayes:
Fixed. The crash was an alloc/kill imbalance in radiate_2_probep (forest/code/Sh_def_kind.f90) introduced by the PTC Pancake commit (1b8a10298); ptc_test now runs to completion in ~2 seconds.

What was happening. The "hang" wasn't a hang: PTC's DA package printed ERROR IN DAINF, DA VECTOR -20 NOT FOUND and then called mypause, which blocks reading stdin. An lldb backtrace (breakpoint on mypause) showed the failure inside ptc_setup_map_with_radiation → radiate_2_probep → powmap → daall1.

Root cause. The commit's rewrite of radiate_2_probep broke TPSA alloc/kill pairing across branches:
- z was killed unconditionally at the end but only alloc'd inside the kind22 branch. For the quads/bends/RF in tiny_ring.bmad, kill(z) ran on an uninitialized real_8, freeing a garbage DA index and corrupting the DA allocation table — the next allocation handed out the bogus index -20 and dainf aborted.
- av was alloc'd unconditionally but killed only in the kind22 branch (a 3-vector leak per call).

Fix (forest/code/Sh_def_kind.f90:25161): alloc z alongside st at the top and kill st, z and av unconditionally at the end, matching the pattern already used in radiate_2_forcep. Rebuilt with CondaCompile.bash; the new generalized_gradient_test from the same commit still passes bit-for-bit, so the pancake feature is unaffected.

One thing needs your judgment. ptc_test output no longer matches output.correct on a few lines, and this is not from my fix — the same commit intentionally rewrote the radiation recoil physics (the transverse kick is now applied via x′ with a curvature factor (1 + b0·x(1)) instead of simple momentum rescaling). Out of tolerance:
- layout-emit: a/b emittances changed from 4.64e-10 / 4.62e-10 to 8.68e-10 / 3.14e-10 (this tilt-coupled ring had near-equal emittances before; they now split substantially), z-emit shifted 0.5%.
- layout-damp (z, 0.5% on a near-zero value) and layout-sigma1 (marginal, 3e-12 vs 2e-12 tolerance).

Everything else agrees within tolerance. If the new radiation physics is expected (worth confirming with David/Étienne), output.correct should be regenerated; I left it untouched.

### DIFF
--- a/forest/code/Sh_def_kind.f90
+++ b/forest/code/Sh_def_kind.f90
@@ -25158,7 +25158,7 @@ endif
         c%delta_rad_out=denf/2+c%delta_rad_out
        endif
 endif
-    call alloc(st)
+    call alloc(st,z)
 
     if(k%TIME) then
        ST=SQRT(1.0_dp+2.0_dp*X(5)/EL%P%beta0+x(5)**2)-1.0_dp
@@ -25207,8 +25207,6 @@ endif
           endif
        ELSEif(el%kind==kind22) then
 
-          call alloc(z)
-
           IF(EL%HE22%P%DIR==1) THEN
              Z= pos*el%l/el%p%nst
           ELSE
@@ -25226,7 +25224,6 @@ endif
              X(4)=(X(4)-EL%P%CHARGE*AV(2))*(1.0_dp+X(5))/(1.0_dp+ST)
              X(4)=X(4)+EL%P%CHARGE*AV(2)
           endif
-          call kill(av,3)
 
        ELSE
  if(k%TIME) then
@@ -25246,13 +25243,13 @@ endif
        ENDIF
     endif
 
-    call kill(st)
+    call kill(st,z)
+    call kill(av,3)
 
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
     p%x=x
     call kill(x)
-           call kill(z)
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
   end subroutine radiate_2_probep


ChristopherMayes:
Claude comparing with PR 2096:

#2096 is the same bug, an equivalent fix, and it goes a bit further than we did. Our local build passes Dave's new baseline perfectly — I checked all 67 data lines of our output.now against the PR's output.correct with the proper ABS/REL tolerances: 0 failures.

Same diagnosis. The PR identifies exactly the alloc/kill mismatch we found in radiate_2_probep: kill(z) moved to the unconditional tail while alloc(z) stayed inside the kind22 branch, and kill(av,3) stranded inside kind22 while alloc(av,3) became unconditional. His DA vector index was -9 where ours was -20 — same table corruption, different garbage value (the PR's failure-chain analysis explains the negative index as an unbounded downward free-slot search after the phantom free desyncs nhole from allvec). It also confirms the crash is release-only, which is why it may not have shown up in debug builds.

Fix shape differs slightly, functionally identical. Dave's is a 2-line swap: kill(z) back inside kind22, kill(av,3) to the tail. Ours allocs z unconditionally with st and kills everything at the tail. Both are balanced; his is the more minimal diff and keeps z's lifetime confined to the branch that uses it. Since both edits touch the same hunks, our local Sh_def_kind.f90 edit will conflict when PR 2096 merges — we should revert ours and take the PR. Say the word and I'll revert it.

Things in the PR we didn't have:
- The ekick => null() initializer on CAV4/CAV4P in a_def_all_kind.inc — a real latent defect (undefined pointer association status) we didn't look for.
- Regenerated baselines for mat6_calc_method_test and long_term_tracking_test, which we never ran, plus ptc_test's.

One caveat on the PR description. The body describes the ptc_test baseline update as just a ~3e-12 layout-sigma1 shift, but the actual diff also bakes in the layout-emit change I flagged yesterday — a/b emittances going from 4.64/4.62e-10 to 8.68/3.14e-10, an ~87% change in the coupled-ring emittance split from the radiate_2 recoil-physics rewrite. The PR's mat6_calc_method_test section does justify the pancake physics change, but the emittance change in ptc_test is accepted silently. Our independent build reproduces his exact new values (agreement to ~1e-5 relative), so it's consistent and deliberate physics from the #2094 resync, not a platform artifact — but if anyone asks why the ptc_test emittances moved, the PR text won't answer it; that might be worth a comment on the PR before it merges.

I fetched the PR as local branch pr2096-check for the comparison; it's harmless to leave or delete._

DavidSagan:
@ChristopherMayes Looks like this PR can be deleted. You agree?

ChristopherMayes:
Yup, close.

PR #2098, opened 2026-07-10
CI: version tag on release + use release tag for Tao

ken-lauer:
## Background

Currently, the released version that Tao reports does not sync up with the latest release tag (and hence the version you would `conda install` as well).
This is because the version bump only happens when @DavidSagan commits and pushes code. Additionally, the reported version from Tao uses a different format than the release tags.

I'd like to propose synchronizing those, and doing it automatically with GitHub Actions.

## Changes in this PR

* Change the `tao_version_mod` to report the git tag
* Have the GitHub actions bot commit to main on every release, a single commit indicating `REL: 2026mmdd` that just includes the version number.
* Fixed permissions for non-release workflows - they don't need write permissions

## Requirements

If you think this is reasonable, @DavidSagan, you'll need to create a personal access token to allow the GitHub Actions bot to include the release commit to `main`.

1. Go to GitHub → Settings (personal) → Developer settings → Personal access tokens → Fine-grained tokens → Generate new token.
  - Resource owner: `bmad-sim`
  - Repository access: Only select repositories (`bmad-ecosystem`)
  - Permissions → Repository permissions → Contents: Read and write (only one required)
  - Set an expiration and note it (**important**, as the release workflow breaks when it expires).
2. Add it as a repo secret to `bmad-ecosystem`
  - Repo Settings → Secrets and variables → Actions → New repository secret, name it exactly `RELEASE_PAT`, paste the token. 

### Miscellaneous

* `update_tao_version.py` is missing from the repo; I assume it's only something you have David


DavidSagan:
@ken-lauer Claude says this. What do you think?

⏺ The short answer

  Don't use a PAT at all. Fine-grained PATs are capped at 366 days, and classic PATs that never expire are a security/maintenance liability (broad scope, tied to your personal account). The
  purpose-built, no-expiry mechanism is a GitHub App: you store the App's private key, which never expires, and the workflow mints a short‑lived (1‑hour) installation token at runtime.

  Recommended: GitHub App token

  1. Create a GitHub App under the org: bmad-sim → Settings → Developer settings → GitHub Apps → New.
    - Repository permissions → Contents: Read and write (that's all you need).
    - Generate a private key (downloads a .pem) — this is the thing that doesn't expire.
    - Note the App ID.
  2. Install the App on bmad-ecosystem (Install App → only that repo).
  3. Store two things in the repo:
    - APP_ID as a repo variable (not even secret).
    - APP_PRIVATE_KEY as a repo secret (paste the .pem contents). No expiry to track.
  4. Change release.yml to mint a token instead of using the PAT:

        - name: Generate app token
          id: app_token
          uses: actions/create-github-app-token@v2
          with:
            app-id: ${{ vars.APP_ID }}
            private-key: ${{ secrets.APP_PRIVATE_KEY }}

        - uses: actions/checkout@v7
          with:
            token: ${{ steps.app_token.outputs.token }}
  
  The rest of the PR's push step (git push origin HEAD:main) stays as-is.

  Important: your main protection will block the push regardless of token

  I checked the current settings on main:

  - Require a pull request before merging is on, and
  - enforce_admins: true — so admins get no exception either.

  That means a direct git push origin HEAD:main from the workflow will be rejected whether you use a PAT (as you) or a GitHub App — classic branch protection has no per-actor bypass. So the
  PAT plan in the PR would likely fail on the first real release.

  The clean fix that pairs with the GitHub App: migrate main from classic branch protection to a repository ruleset, and add your GitHub App to the ruleset's Bypass list. Then the App can
  push the REL: commit while everyone else still goes through PRs — and nothing expires.

ken-lauer:
I haven't tried GitHub Apps before, but I'd be willing to do the adjustments and give it a try - if you handle the admin-related things on your end.

DavidSagan:
I am closing this and will create a new PR using a GitHub App.

PR #2105, opened 2026-07-14
Fix sign error in exact sbend hard-edge fringe transfer matrix

ChristopherMayes:
# Fix sign error in exact sbend hard-edge fringe transfer matrix

## Summary

This PR fixes a sign error in the analytic transfer matrix of the exact (`fringe_type = full`) sbend hard-edge fringe map (`ptc_fringe_dipoler` in `bmad/modules/fringe_mod.f90`). The bug affects **only** the transfer matrix; the tracked orbit is unchanged. The correction is a single term in the energy–time coupling element of the map's Jacobian.

## The bug

In `ptc_fringe_dipoler`, the matrix element `d(3,3) = time_fac / pz` has its derivative with respect to the PTC energy coordinate computed as:

```fortran
d33_dx5 = -time_fac*dpz_dx5/pz**2 - dtime_fac_dx5/pz   ! incorrect
```

Since `d(3,3) = time_fac / pz`, the correct derivative is

```
d/dx5 ( time_fac / pz ) = -time_fac*dpz_dx5/pz**2 + dtime_fac_dx5/pz
```

i.e. the last term should be **added**, not subtracted. (The neighbouring `d13_dx5` and `d23_dx5` terms in the same routine already carry the correct sign; only `d33_dx5` was wrong.)

## The fix

```fortran
d33_dx5 = -time_fac*dpz_dx5/pz**2 + dtime_fac_dx5/pz   ! corrected
```

## Impact / scope

- **Matrix only.** The orbit map does not use this derivative, so tracking results are bit-for-bit unchanged.
- The error was confined to the transfer-matrix term coupling energy to time of flight (`mat6(6,5)` in PTC coordinates) for sbends tracked with the exact hard-edge fringe. Its magnitude scales with `fint * hgap * g`, so it is zero for elements without pole-face field integrals and small (but non-zero) otherwise.
- Downstream quantities derived from the one-turn / element transfer matrix (e.g. chromatic and time-of-flight matrix elements, and anything computed from them) are corrected accordingly.

## Verification

The correct sign was established two independent ways:

1. **Analytic:** direct differentiation of `d(3,3) = time_fac / pz`.
2. **Numerical:** finite-differencing the fringe *orbit* map's arrival-time coordinate with respect to the energy coordinate agrees with the corrected expression and disagrees with the original.

As a further independent check, after the fix the analytic `Bmad_Standard` transfer matrix for a bend with finite `fint`/`hgap` now agrees with the `Tracking` and `Runge_Kutta` matrix methods for the same element (`SBEND5:Bmad_Standard:MatrixRow5` in `mat6_calc_method_test`), whereas before the fix `Bmad_Standard` disagreed with those methods at the level of the bug (~3e-8 for the test lattice).

## Testing

- Full Fortran regression suite passes: **52 passed, 3 skipped** (the skips are pre-existing).
- Exactly one regression reference value changes as a result of the fix: `SBEND5:Bmad_Standard:MatrixRow5` in `regression_tests/mat6_calc_method_test/output.correct`. This reference is updated in this PR.

## Changed files

| File | Change |
| --- | --- |
| `bmad/modules/fringe_mod.f90` | One-line sign fix in `ptc_fringe_dipoler` (plus explanatory comment). |
| `regression_tests/mat6_calc_method_test/output.correct` | Corrected reference value for `SBEND5:Bmad_Standard:MatrixRow5`. |

## Backwards compatibility

Tracking is unaffected. Transfer matrices for exact-fringe sbends with finite `fint`/`hgap` change slightly (they become correct); results derived from those matrices will change at the level noted above.

## Follow-up

A separate, larger change (not part of this PR) reimplements `exact_bend_edge_kick` directly in Bmad phase-space coordinates, removing the PTC coordinate round-trip. That work builds on this fix and will be submitted independently.

---

### Note on authorship / tooling

The sign error was identified and analysed, and this fix and its verification prepared, with the assistance of GitHub Copilot (Claude model) under human direction and review. All changes were validated against the Bmad regression suite and reviewed before submission.


bmad/modules/fringe_mod.f90
```diff
@@ -1266,7 +1266,9 @@ subroutine ptc_fringe_dipoler(X, g_tot, beta0, fint, hgap, particle_at, err_flag
 
   d33_dx2 = -time_fac*dpz_dx2/pz**2
   d33_dx4 = -time_fac*dpz_dx4/pz**2
-  d33_dx5 = -time_fac*dpz_dx5/pz**2-dtime_fac_dx5/pz
+  ! d(3,3) = time_fac/pz, so d/dx5 = -time_fac*dpz_dx5/pz**2 + dtime_fac_dx5/pz.
```
MichaelEhrlichman, line 1269: Do the bugged lined need to be kept?  This is cruft.

bmad/modules/fringe_mod.f90
```diff
@@ -1266,7 +1266,9 @@ subroutine ptc_fringe_dipoler(X, g_tot, beta0, fint, hgap, particle_at, err_flag
...
+  ! The sign of the last term was previously wrong (bug fix).
```
MichaelEhrlichman, line 1270: Perhaps this is not needed in the code base.

MichaelEhrlichman:
It looks like this fixed a simple typo (missing sign) in the code base.  I see that Claude took a derivative of d33 to verify the mistake, though I would still verify this against another source like a document or another code base.

MichaelEhrlichman (approved):

DavidSagan:
Verified by comparing to RK tracking.

PR #2106, opened 2026-07-14
Track the exact sbend hard edge in Bmad coordinates

ChristopherMayes:
# Track the exact sbend hard edge in Bmad coordinates

## Summary

This PR reimplements `exact_bend_edge_kick` (the exact, `fringe_type = full` sbend hard-edge fringe map) so that it tracks directly in Bmad phase-space coordinates, removing the round-trip conversion to and from PTC coordinates. The map is unchanged physically; the new code reproduces the previous results to machine precision and runs faster. The Bmad-coordinate formulas are also added to the manual.

This PR is stacked on the fringe-matrix sign-fix PR (`bugfix-exact-bend-fringe-matrix`); please review/merge that one first. Its base should be retargeted to `main` once the bugfix merges.

## Motivation

The previous implementation converted the orbit from Bmad coordinates to PTC coordinates (`vec_bmad_to_ptc`), applied three PTC-derived maps (`ptc_wedger`, `ptc_rot_xz`, `ptc_fringe_dipoler`), and converted back (`vec_ptc_to_bmad`). The PTC round-trip is unnecessary: the four transverse coordinates are identical in the two conventions, and only the energy/longitudinal pair is reparameterized. Working directly in Bmad coordinates removes the two nonlinear conversions (and their Jacobians), simplifies the data flow, and is faster.

## What changed

- New Bmad-native helper routines `bmad_wedger`, `bmad_rot_xz`, and `bmad_fringe_dipoler` compute both the orbit and the analytic transfer matrix directly in Bmad phase space.
- `exact_bend_edge_kick` now composes these helpers with no coordinate conversion.
- The previous PTC-coordinate implementation is retained (renamed `exact_bend_edge_kick_ptc`, with its `ptc_*` helpers) purely as a reference for the regression test and benchmark. It, and the `ptc_*` helpers, can be removed in a later cleanup once there is confidence in the new code.
- The manual section "Exact Dipole Hard Edge Fringe Map" (`bmad/doc/fringe-fields.tex`) now gives the Bmad-coordinate wedge and hard-edge fringe transport formulas.
- A standalone derivation note is added under the non-built `bmad/doc/notes/` folder.

## Correctness

The transfer-matrix formulas were derived symbolically and checked against finite differences before being written into the code. The new regression test compares the Bmad-native routine against the PTC reference over a scan of starting orbits, both edges, and both time directions. Agreement (for the test lattice):

- orbit: max difference ~5e-17
- transfer matrix: max difference ~4e-16
- arrival time: max difference ~2e-25
- symplectic error of the new matrix: ~6e-16

The full Fortran regression suite passes (52 passed, 3 pre-existing skips; 53 with the new test). One reference file (`tracking_method_test`) is updated for round-off-level orbit changes that come from evaluating the map without the PTC round-trip.

## Performance

Timing the two implementations back-to-back (2,000,000 calls each; run `exact_bend_edge_test bench`):

- orbit only: Bmad-native ~147 ns/call vs PTC reference ~205 ns/call (~1.4x faster)
- with transfer matrix: Bmad-native ~615 ns/call vs PTC reference ~947 ns/call (~1.55x faster)

## New test and benchmark

`regression_tests/exact_bend_edge_test/` contains a standalone program that (i) checks the Bmad-native orbit, arrival time, and transfer matrix against `exact_bend_edge_kick_ptc`, (ii) checks the symplecticity of the new matrix, and (iii) benchmarks both implementations. The benchmark is gated behind a command-line argument, so it does not run under the regression harness.

## Changed files

| File | Change |
| --- | --- |
| `bmad/modules/fringe_mod.f90` | Bmad-native `exact_bend_edge_kick` plus `bmad_wedger`, `bmad_rot_xz`, `bmad_fringe_dipoler`; old routine kept as `exact_bend_edge_kick_ptc`. |
| `bmad/doc/fringe-fields.tex` | Bmad-coordinate formulas for the exact dipole hard edge fringe map. |
| `bmad/doc/notes/exact-bend-edge-bmad-coords.tex` | Standalone derivation note (non-built). |
| `regression_tests/exact_bend_edge_test/` | New regression test and benchmark, plus registration. |
| `regression_tests/tracking_method_test/output.correct` | Reference updated for round-off-level orbit changes. |

## Backwards compatibility

The map is physically identical; results agree with the previous implementation to the precision quoted above. No user-facing behavior changes.

## Follow-up

A later cleanup can remove `exact_bend_edge_kick_ptc`, the `ptc_*` helpers, and the `vec_bmad_to_ptc` / `vec_ptc_to_bmad` conversions, and switch the regression test from a new-vs-old comparison to a stored-reference snapshot.

---

### Note on authorship / tooling

The reimplementation, its symbolic and numerical verification, the regression test and benchmark, and the manual documentation were prepared with the assistance of GitHub Copilot (Claude model) under human direction and review. All changes were validated against the Bmad regression suite and reviewed before submission.


ChristopherMayes:
# Review

**Verdict: correct and well-executed — recommend merge after #2105, with one minor robustness suggestion.** Verified three ways: line-by-line derivative checking against hand differentiation, comparison of every transverse formula against the PTC originals, and building + running the full regression suite on this branch (macOS/arm64, gfortran via conda).

## Correctness — verified

- **The coordinate-change argument is sound.** The four transverse coordinates are identical between the two conventions; only `(z, p_z) ↔ (τ, p_τ)` differ, and since the edge fields are static-magnetic, β is invariant, so every PTC time update maps to `Δz = −β·Δτ` (which is `−(1+p_z)·L` for the wedges since `β·ℰ = 1+p_z`). The transverse formulas in `bmad_wedger`/`bmad_rot_xz`/`bmad_fringe_dipoler` are verbatim copies of `ptc_wedger`/`ptc_rot_xz`/`ptc_fringe_dipoler`, and the z updates follow this rule exactly.
- **The matrix derivatives check out.** Every term was hand-checked: the wedge Jacobian (including the `arcsin` chain rules through `p_t`), the fringe's `dfi0/dco1/dco2/dfi1/dfi2/dfi3` derivative cascade, the `yc/m33/tauc` structure, and the delta-column terms including the `dβ/dδ = (1/β₀²−1)/ℰ³` contribution to `mat6(5,6)`. The `d33` column carries the corrected sign from #2105.
- **Driver parity.** The new `exact_bend_edge_kick` reproduces the old driver exactly: same `e1/e2`, `fint/hgap` vs `fintx/hgapx` selection via `physical_ele_end`, same `time_dir` handling, same map composition order per edge, same `lost_pz$` error paths. The `orb%t` update `−Δz/(β·c)` is algebraically identical to the old `+Δτ/c`. The only production call site is `bend_edge_kick` in the same module, and the signature is unchanged.
- **Manual matches code.** The `fringe-fields.tex` formulas are consistent with the implementation — the manual defines the `d` matrix's third column *without* the ℰ factor and compensates with `(1+p_z)` in the z̄ equation, which is consistent since `β·ℰ = 1+p_z`.

## Build & test — reproduced independently

- Branch builds cleanly.
- `exact_bend_edge_test` reproduces the PR's numbers exactly: orbit diff 5.6e-17, matrix diff 4.4e-16, time diff 2.1e-25, symplectic error 6.0e-16.
- Benchmark: 120 vs 164 ns/call (orbit) and 415 vs 646 ns/call (matrix) — a 1.4×/1.6× speedup, slightly better than claimed.
- Full regression suite: 10,688 tests, **0 failures** (exactly 4 more than base — the new test). The single "flow failure" is the pre-existing `bookkeeper_test` harness parse issue from #2104 on main (`run_tests.py` cannot parse the unquoted `"Ramper-loc-1" STR 26 0` lines), unrelated to this PR.
- Confirmed the #2105 bugfix commit is an ancestor of this branch.

## Suggestions

1. **(Minor, robustness) Restore the small positive floor in `bmad_wedger`.** `ptc_wedger` guards `radix < 1e-10` before both `sqrt(pt²)` and `sqrt(pzs²)`; the new code only checks `<= 0`. For near-grazing rays with `psn²` between 0 and 1e-10, the matrix branch divides by `psn` and by `sbin = sqrt(1−(pxn/pt)²) → 0`, which can overflow or NaN where the old code cleanly flagged `lost_pz$`. A `1e-10` floor keeps behavioral parity.
2. **(Nit) Unused `beta0` arguments.** `bmad_rot_xz` never uses `beta0`, and `bmad_wedger` only forwards it — a natural consequence of the `Δz = −(1+p_z)L` simplification. Fine for signature uniformity, but could be dropped in the planned cleanup along with the unused `r_name` constants.
3. **(Nit) Tolerance margin.** The updated `tracking_method_test` reference (`RBEND4-Anti_D`) moved by 4e-13 — exactly its `ABS 4e-13` tolerance. If CI platforms show different round-off, this line is the one most likely to flicker; loosening it slightly would be cheap insurance.

Two things the rewrite quietly *improved* over the old code, worth keeping: the old `ptc_fringe_dipoler` had `if (fac == 0) return` (negative `fac` would NaN) — the new code correctly uses `<= 0`; and the old matrix branch computed `sqrt(1−2·bb·y)` *before* the negativity check, while the new code checks first.

## Test coverage

Good: 6 phase-space starts (including on-axis and `p_z = ±0.2…0.5`), both edges, both time directions, matrix + orbit + arrival time + symplecticity, at a moderately relativistic energy (1 MeV electron, β₀ ≈ 0.86) where the β-dependence actually matters. The one structural limitation is inherent to new-vs-old testing: a bug shared by both implementations wouldn't be caught — but the independent symplecticity check and the suite-wide `Bmad_Standard`-vs-`Tracking` matrix comparisons mitigate that. The planned follow-up switch to a stored-reference snapshot (once `exact_bend_edge_kick_ptc` is removed) is the right call.

---

*Review prepared with the assistance of Claude Code under human direction; all build and test results were reproduced locally.*

ChristopherMayes:
Thanks for the thorough review. Addressed in 0b1f38bad:

1. **Robustness floor (done).** Restored the `1e-10` positive floor in `bmad_wedger` on the `pt` and `psn` radicands, matching `ptc_wedger`, so near-grazing rays that would otherwise divide by `psn` / `sbin` cleanly hit `lost_pz$` instead of overflowing to NaN. Kept the `ps` guard at `<= 0` to mirror ptc's `fac < 0` check on the longitudinal-momentum radical.
2. **Unused `beta0` args (deferred).** Left as-is for signature uniformity; will drop these (and the unused `r_name` constants) in the planned cleanup that also removes `exact_bend_edge_kick_ptc`, the `ptc_*` helpers, and the `vec_*_to_*` conversions.
3. **Tolerance margin (done).** Loosened the `RBEND4-Anti_D: Bmad_Standard` line in `tracking_method_test/output.correct` from `ABS 4e-13` to `1e-12` so it won't flicker across CI platforms.

Rebuilt cleanly (macOS/arm64, gfortran via conda) and both affected tests pass (`exact_bend_edge`, `tracking_method`).

PR #2117, opened 2026-07-31
STY: fix and reformat .init files for libgfortran 16.x compatibility

ken-lauer:
## Changes

* Reformats `tao.init` and `tao_plot.init` files with `latform`'s new namelist support.
* Fixes unquoted string fields (most common in `universe = n` which seems like it should be an integer)
* Normalizes logicals: `.true.`/`.false.` and currently accepted equivalents become `T`/`F`
* Aligns equals signs and comments in namelist groups
* Uses string constants for colors and other curve enums instead of their integer equivalents for consistency

## Note

I'd recommend viewing this PR diff with whitespace changes hidden:

image

## Motivation and Context
The recent release of libgfortran 16.0 changed a key aspect of namelist input file evaluation (ref: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=123012). We've just now seen the effects of this due to conda-forge upgrading.

Derived type fields that are of base type `character` (i.e., strings) must have quotes around them, or the namelist parser will cause the process to exit with an error. For example:

```
&tao_d2_data
        d2_data%name = 'L0B'
        universe = 1
        n_d1_data = 1
/
```
Since `universe` is actually a `character(16)`, this is now invalid as of libgfortran 16. It causes `tao` to exit with this error:

```
At line 107 of file /home/conda/feedstock_root/build_artifacts/bmad_1785112804039/work/tao/code/tao_init_data_mod.f90 (unit = 30, file = 'tao.init')
Fortran runtime error: Missing quote while reading item ...
```

There are numerous other such fields in existing `.init` files that needed fixing.

## Additional testing

I made sure that each `tao.init` at least didn't crash Tao. There are a number of issues that need resolving, but there was only one outright segfault that I believe is unrelated to these changes (`./bmad-doc/lattices/cbeta/model/sub/in`)

Consider using a script like this to try for yourself.

```bash
#!/usr/bin/env bash

readarray -d '' -t init_paths < <(find . -type f -name tao.init -print0)

for path in "${init_paths[@]}"; do
	echo "---------------------"
	echo "$path"
	echo
	echo
	parent=$(dirname $path)
	(cd $parent && tao -noplot -command 'quit' || echo "failed: $path")
done
```


DavidSagan:
@ken-lauer Why is this in draft mode?

ken-lauer:
It should be ready now.

I'm not sure if you'd care to merge it in or handle it yourself; so at least take it as a basis for what needs to be fixed.

PR #2145, opened 2026-08-13
Doc update.

DavidSagan:


DavidSagan:
### The `ubuntu-single-core` CI failure is not caused by this PR

This PR only touches `bmad/doc/lattice-file.tex` and a PDF. The same job has been failing on `main` since run [31736368998](https://github.com/bmad-sim/bmad-ecosystem/actions/runs/31736368998) (13-Aug 19:32) and on every branch pushed since. The last green run was 12-Aug 19:53.

**Cause: a conda-forge toolchain update, not a source change.**

The failure is at the final link of `tao`:

```
ld: warning: libstdc++.so.6, needed by .../production/lib/libsim_utils.so, not found (try using -rpath or -rpath-link)
ld: libsim_utils.so: undefined reference to `std::cout@GLIBCXX_3.4'   [+11 more]
collect2: error: ld returned 1 exit status
```

`libsim_utils.so` contains C++ (`Faddeeva.cc`, `opti.cpp`, `read_dir.cpp`, `cpp_fortran_utils.cpp`), so it carries a `DT_NEEDED` on `libstdc++.so.6`. When `ld` resolves the *indirect* dependencies of a shared library on the link line it searches `-rpath-link` / `-rpath` / `DT_RUNPATH`, but **not** `-L`. Nothing was supplying that path.

Diffing the conda packages between the last green job (`94242705306`) and the failing one (`94631259808`):

| | green (12-Aug) | failing (13-Aug on) |
|---|---|---|
| `compilers` | 1.11.0 | **2.0.0** |
| `gcc` / `gfortran` / `gxx` | 14.4.0 | **15.3.0** |
| `conda-gcc-specs` | 14.4.0 | **gone** |
| `gcc_linux-64`, `gxx_linux-64`, `gfortran_linux-64`, `binutils_linux-64` | present | **gone** |

conda-forge `compilers 2.0.0` dropped `conda-gcc-specs` and the `*_linux-64` activation packages, which were what implicitly injected the conda lib dir into every link. The build was relying on that.

**Why the other two jobs still pass:**

* `ubuntu-openmp` links through `mpifort`, whose flags are (from the log) `-L$CONDA_PREFIX/lib -Wl,--allow-shlib-undefined -Wl,-rpath -Wl,$CONDA_PREFIX/lib ...` — either half of that hides the problem. `ubuntu-single-core` links with bare `gfortran` and `ACC_LINK_FLAGS = -lX11 -ldl`.
* macOS records absolute install names, so indirect deps need no link-time search path.

**Fix** (in progress, separate PR): supply the conda lib dir explicitly rather than relying on the toolchain to do it.

* `util/acc_build_common` — the conda `-L`/`-I`/`-rpath` block lived inside `func_set_openmp_flags`, i.e. it only ran for the one configuration that didn't need it. Hoisted into a `func_set_conda_flags` that runs for every conda build, and `-Wl,-rpath-link` added.
* `util/Master.cmake` — the CMake build (where the `tao` link actually happens) does not read `LDFLAGS`, so `-L`/`-rpath`/`-rpath-link` for `$ACC_CONDA_PATH/lib` are appended to `ACC_LINK_FLAGS` and `CMAKE_SHARED_LINKER_FLAGS` for non-Darwin conda builds. The latter also gives the `.so` files a `DT_RUNPATH`, which fixes the problem at its source.

Note `tao/cmake.tao` already does `link_directories($ENV{ACC_CONDA_PATH}/lib)`, but `-L` alone cannot fix this — hence the `-rpath-link`.

This PR can be merged once CI is fixed; nothing here needs changing.

DavidSagan:
Correction to my note above: rather than a separate PR, the fix is now pushed onto this branch as 2718dfc74, so this PR's own CI run is the test of it.

* `util/acc_build_common` — conda `-L`/`-I`/`-rpath` flags hoisted out of `func_set_openmp_flags` into `func_set_conda_flags`, which runs for every conda build, plus `-Wl,-rpath-link`.
* `util/Master.cmake` — the CMake build does not read `LDFLAGS`, so the same flags are added to `ACC_LINK_FLAGS` and `CMAKE_SHARED_LINKER_FLAGS` for non-Darwin conda builds. The latter gives the `.so` files a `DT_RUNPATH`, fixing it at the source.

If you would rather keep this PR to the doc change alone, the commit is self-contained and can be cherry-picked out onto its own branch.

DavidSagan:
@ChristopherMayes @ken-lauer Hi Ken: Do you approve of this revision?

ken-lauer (approved):
Since it builds here, I approve.

Once it gets into a release and onto a conda-forge tag, it may need revisiting - I'll let you know then.

Issue #1960, opened 2026-03-09
`pytao.so` Executable Stack Flag


electronsandstuff:
**Describe the bug**
Bmad's build system sets the executable stack flag in shared libraries it generates. This causes problems on some modern systems where software with the flag set are forbidden for security reasons. As an example, launching `tao` inside of a docker container based on the image `python:3.10-slim` (Debian-based) causes the following error.

```
server  | OSError: /opt/conda/envs/example/lib/libtao.so: cannot enable executable stack as shared object requires: Invalid argument
```

My understanding is that this is related to Fortran nested functions and the use of stack [trampolines](https://gcc.gnu.org/onlinedocs/gccint/Trampolines.html). Moving away would require refactoring the code to avoid the use of nested functions. This is a big request and I am filing this issue to raise it to people's attention and discuss possible paths forward.

As a note if work is done in this area, `gfortran` includes an option to warn on the generation of tramplines (`-Wtrampolines`)

ken-lauer:
This is exactly the issue that I worked around in https://github.com/bmad-sim/bmad-ecosystem/pull/1891.

MacOS doesn't support an executable stack, so the trampolines were heap-allocated leading to massive slowdowns. 
I'd recommend against just "fixing"/patching over this this with compiler flags as there could be similarly severe performance implications, and instead attempting to refactor as in the above PR by moving up such nested functions.

DavidSagan:
A refactor is doable. Here is a list of what must be changed:
```
sim_utils/math/fourier_mod.f90
bmad/code/attribute_bookkeeper.f90
synrad3d_wall_to_synrad_walls_mod.f90
synrad3d_track_mod.f90
wake_fit/wake_fit.f90
generalized_gradient_fit/gg_fit_mod.f90
```
For Tao only the first two are needed to be modified.

jsberg-bnl:
This is really a gfortran problem combined with the fact that you are probably running a very cutting-edge Linux. Summarizing the fixes:

- At build time, if building with gfortran version 14 or later, set `FFLAGS=-ftrampoline-impl=heap` in the environment. @ken-lauer, is the Conda build with a sufficiently modern gcc on Linux? You could just do that and fix the problem.
[...]

PR #2153, opened 2026-08-17
Replace Fortran namelist reads in tao/code with a hand-written reader

DavidSagan:
Removes every `namelist` statement from `tao/code` and replaces the reads with a hand-written reader.

**Back compatibility was treated as a hard requirement.** The input file syntax is unchanged; only the reader is replaced, as a strict superset of what a namelist read accepts. Existing input files behave exactly as before.

## What this buys

* Error messages name the file and the line, and one bad item is reported as one bad item instead of aborting the whole group.
* Array sizes no longer have to be fixed at compile time by the declaration of the namelist variable. The number of datums in a `d1_data` and the number of variables in a `v1_var` are now counted from the file rather than capped at a guessed constant.
* No dependence on how a particular compiler chose to interpret the standard.

## How it is put together

| File | Role |
| --- | --- |
| `tao/code/tao_nml_mod.f90` | The reader. Gathers one group into a list of `(name, value, file, line)` items that the call site applies. |
| `tao/code/tao_attrib_ptr_mod.f90` | Runtime. Turns an item value into a stored value. |
| `tao/code/tao_attrib_resolve_mod.f90` | Generated. Maps a component name, or a flattened slot index for positional assignment, to a pointer into the struct. 49 structs, 619 components. |
| `tao/scripts/generate_attrib_tables.py` | The generator. `--check` mode is run by `test_generated.py` so the tables cannot drift from the struct definitions. |

Name-to-variable binding stays at the call site as a `select case`. A generic runtime binding would need either polymorphic wrappers for all 49 structs or internal procedure callbacks, and the latter would reintroduce the gfortran trampoline problem of #1960.

## Deliberate differences from a namelist read

Three, all of which only ever accept more than before:

* A numeric scalar value may be an expression, so `n_turns = 40 + 60` now works.
* A group with no terminating `/` ends at end of file. A namelist read aborts, or with `iostat` present discards the group silently.
* A blank value is an error rather than being silently ignored.

Everything else is deliberately identical, including several cases a plain list-directed read gets wrong:

* A short value list for an array leaves the remaining elements unchanged. `d_orb = 1e-6, 2e-6` sets two of six.
* A null value leaves that element unchanged. `1.0,,3.0` does not zero the middle element.
* More values than the array holds is an error. `d_orb = 7*1e-6` is rejected.
* A value that cannot be decoded is an error even in the middle of a list. `d_orb = 1e-6, junk, 3e-6` reads exactly as many values as `d_orb = 1e-6` does, so this needs the values to be counted separately from the read.

Subscript ranges, open-ended ranges, repeat counts, and tabs all behave as they did.

## Expressions

The evaluator is `tao_evaluate_expression`, so anything it knows works, named constants included:

```
global%delta_e_chrom        = 2 * pi
global%n_opti_cycles        = 40 + 60
bmad_com%max_aperture_limit = sqrt(m_electron) * 1e-3
```

Two limits worth knowing, both of which follow from the file syntax being kept as is rather than from a choice made here:

* **A `/` cannot appear in a value**, since `/` ends the group. Write `c_light * 1e-9`, not `c_light / 1e9`. This is not new — a namelist read ends the group at the `/` too — but it is silent, so it is worth stating.
* **Only a scalar is evaluated.** A value list for an array is not, so `d_orb = 1e-6, 2*pi*1e-6` is an error rather than being quietly taken as a repeat count.

An expression is only tried where a namelist read would have failed, so `rep = 3*1.5` for an array is still a repeat count. For a scalar there is no such ambiguity to preserve, since a repeat count supplies more than one value and a namelist read rejects more than one value for a scalar — so `2*pi` is a multiply.

## Verification

The full Tao state (`show global`, `lattice`, `data`, `var`, `plot`, and so on) was diffed between the old and new binaries over the `tao_examples` in bmad-doc. **No differences.**

Two new regression tests:

* `tao_nml_test` reads the same file twice, once with a real namelist read and once with the new reader, and compares the results.
* `tao_attrib_test` covers the resolver and the value setter.

## Notes for review

* `tao_open_scratch_file` now has no callers. Left in place since it is public; removing it is a separate decision.
* Include / `call, file =` support is not implemented. It needs the reader to own a file stack because the active unit changes mid-read, and it is new syntax nobody has asked for.
* Four pre-existing bugs turned up while testing and are **not** fixed here, all confirmed present on unmodified `main`: `set dynamic_aperture` segfaults when `u%dynamic_aperture%pz` is unallocated; `write namelist -variable` crashes partway and truncates its output; `set space_charge_com sigma_cut` is a dead alias (the component is `lsc_sigma_cutoff`); and `tao_set_mod.f90` has a missing comma after `'prompt_color'&`.


DavidSagan:
Too fragile and too complicated.

PR #2186, opened 2026-09-11
Beginnings of a pure CMake build system

jsberg-bnl:
Currently has many limitations, but it should be capable of:
- Building and running tao
- Building executables against the bmad libraries

Currently not working
- pgplot as the plot library
- conda build
- tao help
- listf/getf
- mpi
- tests

The bottom of README.md describes proper usage.


DavidSagan (commented):
Review of the new CMake build. It was built from a clean checkout of this branch on macOS (arm64, gfortran 15.2, CMake 3.31) against an existing dependency prefix, then installed; the installed `tao` was run and the README's downstream `find_package(Bmad)` example was built.

The direction is right. Replacing the environment-variable-driven `Master.cmake` with real per-target dependencies, out-of-source builds and an exported `Bmad::` package is a big improvement, and `find_package(Bmad)` + `target_link_libraries(... Bmad::bmad)` for user programs worked once the dependency paths were set. Keeping the legacy path in the same files is also the right call for a transition. As it stands, though, the build fails on macOS, the default build is unoptimized, and on the test machine the installed `tao` silently ran the *old* libraries.

## Confirmed by building

1. **Doesn't link on macOS.** `libsim_utils` fails to link with undefined `_dgesdd_`, `_dpotrf_`, `_dpotrs_`, `_ddot_`, `_dnrm2_`, … `sim_utils` calls BLAS/LAPACK directly (`geodesic_lm`, etc.), but only `lapack95` is linked. On Linux, `ld` lets a shared library carry unresolved symbols and resolves them later through `liblapack95.so`'s dependencies, so this likely only shows up on macOS. Fix: link LAPACK/BLAS through the `lapack95::lapack95` target, e.g.
   ```cmake
   target_link_libraries(lapack95::lapack95 INTERFACE LAPACK::LAPACK)
   ```
   Caveat: a plain `find_package(LAPACK)` on macOS picks Accelerate, which puts two LAPACKs in one process since `lapack95` was built against reference LAPACK. Better to use the dependency prefix's `lapack.pc` (`pkg_check_modules(... IMPORTED_TARGET lapack)`) or set `BLA_VENDOR`.

2. **macOS rpath typo.** `CMakeLists.txt` sets `"@loader_path../lib"` — missing the `/`. Should be `"@loader_path/../lib"`.

3. **Installed `tao` loads the wrong libraries.** Because of #2, the installed `tao` fell back to the legacy `libtao`/`libbmad` that live alongside the dependencies in the prefix (that directory is on the rpath via `CMAKE_INSTALL_RPATH_USE_LINK_PATH`). Verified with `DYLD_PRINT_LIBRARIES`. It looks like it works, but it's running the old code. **Fixing the typo is not enough:** the link-path rpath entries come *before* `@loader_path/../lib`, so an older Bmad build in `CMAKE_PREFIX_PATH` still shadows the new one. That will be a common setup during the switchover. The rpath ordering needs to put the package's own lib dir first.
   Related: `CMAKE_BUILD_WITH_INSTALL_PATH` means binaries in the build tree can't run, which will get in the way once tests/`ctest` are added.

4. **Default build is unoptimized.** The README doesn't mention `CMAKE_BUILD_TYPE`, and when it is empty the compile gets no `-O` flag at all. The `bmad-doc/tao_examples/dynamic_aperture` example took **15.5 s vs 6.4 s** with the current `-O2` production libraries. Suggest defaulting the build type when none is given, with `-O2` set explicitly to match the existing build (CMake's stock `Release` for gfortran is `-O3`, which could shift regression-test results):
   ```cmake
   if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type" FORCE)
   endif()
   ```
   The Debug config should probably also carry the flags `mkd` uses today (`-fbounds-check -Wuninitialized`, …).

5. **Static HDF5 linked into shared `libbmad`.** The HDF5 block copied from `Master.cmake` tests `ENABLE_SHARED`/`ENABLE_SHARED_ONLY`, which are never defined in the new build, so it prefers `hdf5_hl_fortran-static` (the `.a` files are on the `libbmad` link line). The static target is also exported in `Bmad::bmad`'s interface, so a user program that also uses HDF5 ends up with two copies.

6. **`-DENABLE_OPENMP=ON` breaks downstream `find_package(Bmad)`.** The exported `sim_utils` target references `OpenMP::OpenMP_Fortran` and a bare `fftw3_omp`, but `BmadConfig.cmake.in` never calls `find_dependency(OpenMP)`, so the consumer's configure errors out with a missing target.

7. **Minimum CMake version is too low.** `cmake_minimum_required(VERSION 3.14)`, but the `Fortran_PREPROCESS` source property needs 3.18. On 3.14–3.17 it's silently ignored and the preprocessed files fail.

8. **Fragile dependency lookup** (both needed workarounds just to configure):
   - `find_path(lapack95_inc NAMES f95_lapack.mod HINTS "${CMAKE_PREFIX_PATH}")` doesn't search `modules/` (where the existing layout puts `f95_lapack.mod`). Add `PATH_SUFFIXES modules include`.
   - HDF5's own CMake config references `ZLIB::ZLIB` without finding it, so configure fails unless `find_package(ZLIB)` is called first. That's an HDF5 packaging flaw, but the build (and `BmadConfig.cmake`) should work around it.

## Design suggestions

- **Detect the legacy build without env vars.** The `ACC_SET_GMAKE_JOBS` + `GMAKE` test works only because `GMAKE` happens to be exported just inside the `mk` scripts. `if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)` is exact: the legacy system configures each package as its own top-level project, the new one uses them as subdirectories. Moving the legacy content into e.g. `legacy.cmake` and `include()`-ing it would also keep the old logic byte-for-byte unchanged and avoid the re-indented diff. (`cmake_minimum_required()`/`project()` must stay literally in the top-level `CMakeLists.txt`.)
- **Find dependencies in one place.** The lookups in `sim_utils/CMakeLists.txt` and `bmad/CMakeLists.txt` are duplicated in `BmadConfig.cmake.in` and have already drifted (HDF5 looked up without `COMPONENTS` in the config; OpenMP missing). A shared `cmake/BmadDependencies.cmake` included from both would prevent that.
- **Simplify HDF5.** With a minimum of CMake ≥ 3.19 (`FindHDF5` provides `hdf5::hdf5_fortran` / `hdf5::hdf5_hl_fortran` imported targets), most of the ~90 copied lines can go.
- **Build options:**
  - Put OpenMP on `bmad_options` rather than only on `sim_utils`, so `forest` is compiled with it too, as it is today.
  - Guard the GNU-only flags (`-fdollar-ok -ffree-line-length-none`) with `$<Fortran_COMPILER_ID:GNU>`; otherwise ifx breaks, including for downstream users who inherit them through `Bmad::bmad_options`.
  - Make the plot package a cache option (`plplot`/`pgplot`/`none`) and set `CESR_PLPLOT` etc. with `target_compile_definitions` rather than on two individual files.
  - Preprocess the whole target (what the legacy `-cpp` does) instead of listing files; `multiparticle/envelope_mod.f90` is in the list but has no preprocessor directives.
- **Smaller points:**
  - `file(GLOB ... CONFIGURE_DEPENDS)` so newly added `.f90` files are picked up without re-running cmake by hand.
  - Add `project(Bmad VERSION ...)` and a `write_basic_package_version_file()`.
  - `BmadConfig.cmake.in` bakes the build machine's `CMAKE_PREFIX_PATH` into the installed config, so the install can't be relocated.
  - Installing all sources into `src/` adds ~12 MB — presumably for getf/listf; worth stating the intent.
  - `tao/CMakeLists.txt` has `bmad` in two `target_link_libraries` calls.

## Not yet covered (beyond the list in the PR description)

`cpp_bmad_interface_test`, `util_programs`, `lux`, hooking `regression_tests` into `ctest`, and a CI job running the new build on both Linux and macOS — which would have caught #1–#3.

With just the LAPACK and rpath fixes applied, the whole tree built, `tao` ran correctly on the new libraries (forced via `DYLD_LIBRARY_PATH`), and there were no module-name collisions in the merged install `modules/` directory.

DavidSagan:
@jsberg-bnl If there is anything Claude wrote that you don't understand, just post a comment below and I can get an answer. If you want Claude to formulate script changes for your review, again just post a comment.

jsberg-bnl:
OK, this is probably premature, I rushed this and didn't really test it anywhere except Arch. I'll have a look at MacOS in particular. It looks like Claude is getting a couple of things wrong, but I'll have to see what happens when I try it.

PR #2188, opened 2026-09-17
Add tab completion to Tao readline interface

ken-lauer:
## Changes

This is a minimally invasive way of adding tab completion to Tao sessions. Currently, Tao just uses readline's default which auto-completes filenames.

* Centralizes commands/subcommands/flags as best as possible for reuse
* Adds readline C shim to handle calling the Fortran completion helpers
* Adds a `pipe` command so PyTao can benefit too (downstream PyTao will get a PR if this is accepted of course; PyTao tests will fail solely on that until then)

The logic in `tao/code/tao_completion_mod.f90` seems generally good in my testing. Given the nature of the `switch` statements in Tao's command parsing, I'm not sure there's much of a way around duplicating those subcommand names in the completion module - at least not without major refactors which no one should want to see.

## Screen recording

Typing a bit slowly to show how tab completion works in a context-sensitive way:

https://github.com/user-attachments/assets/c7f75764-d3ba-41fd-bdca-825be6dadaac


DavidSagan (commented):
Read through the whole diff, compiled the new sources against installed modules, and looked at the CI logs. Nice work overall — below is what I found.

## What's solid

I mechanically compared every list this PR hoists against what it removes: all 27 converted `show` switch lists, the `set` target list, the `show <what>` list, and the 113-entry `pipe` subcommand list are **identical** to the literals they replace. Since `match_word`/`tao_next_switch` trim names, widening the declared length to `character(28)` is behavior-neutral. Only `'#format'` remains a literal, correctly so since it's a `#` switch.

The C shim compiles clean under `-Wall -Wextra -Wwrite-strings`. The `matches` array ownership contract is right — NULL termination is set correctly even on a partial malloc failure, and the `k + lt + 2 > buf_size` bound genuinely cannot overflow. The new Fortran modules compile clean as well (I built `tao_struct` + `tao_command_names_mod` + `tao_completion_mod`) apart from the truncation warnings noted below. The `word_start - 1 /= istart` bail-out is a careful touch.

## Blocking

**1. Registering the completer hijacks readline for every embedder, including in-process PyTao.**

`tao_register_completion()` is called unconditionally from `tao_top_level`, and `tao_c_interface_mod.f90:108` (`tao_c_init_tao`) routes through `tao_top_level`. So `readline_set_completion_fn` runs in every PyTao process, not just the interactive program, and it permanently overwrites two globals: `rl_attempted_completion_function` and `rl_completer_word_break_characters = " \t"`. On Linux, CPython's `readline` module is normally GNU readline in the same process, so `import pytao; Tao(...)` from IPython would replace IPython's completer and break its word splitting for the rest of the session. Please gate registration on Tao actually owning the prompt, and save/restore the previous values.

**2. Filename completion disappears everywhere the engine does not say `FILE`.**

The shim sets `rl_attempted_completion_over = 1` even when `n == 0`. Only `call` and `read` at `n_words == 1` yield `FILE`, so `write <TAB>`, `spawn <TAB>`, `pipe -write <TAB>`, and every position after a value-taking switch now offer nothing — where today readline's default gives filenames. That is a regression for users who rely on it. Returning `-1` for `context == 'NONE'`, and reserving `0` for "recognized position, no candidates", preserves the old behavior at no cost.

**3. `pipe complete`'s output shape does not match its declared return type.**

CI shows `test_complete_1` failing with `Failed to parse output from command 'pipe complete "sho"': not enough values to unpack (expected 3, got 2)`. That is PyTao's *generic* `string_list` parser rejecting the format — every other `string_list` command emits uniform `a;b;c` lines, whereas this one emits a two-field header followed by bare lines. The PR description attributes the pytao failure to the unmerged downstream PR, but this part is not that. Either conform to the convention or declare a return type that fits.

## Correctness

**4. Zero-size switch arrays are an out-of-bounds read, not an error.** `tao_switches_for` returns `switches(0)` for an unknown key, and `tao_next_switch` does `switch_list(1)(1:1)` as its first statement. Before this PR a bad list was a compile error; now a mistyped key is silent undefined behavior — most relevant for the computed key `tao_switches_for('show ' // trim(show_what))`. Worth a hard failure on unknown key, or a `size(switch_list) == 0` guard in `tao_next_switch`.

**5. The namelist scraping is compiler-format dependent.** gfortran writes ` GLOBAL%N_OPTI_CYCLES=20 ,` (I checked), so the heuristic works there. Intel pads the name with spaces before `=`, which trips `index(nml_line(1:ix2), ' ') /= 0` and hits the `cycle`, giving **zero candidates for every `set <struct> <TAB>`**. ifort is a supported compiler (`util/Master.cmake:100`). Taking the name as everything after `%` up to the first of `' '`, `'='`, `'('` — instead of rejecting on any space — fixes it.

**6. Truncation at `max_matches$ = 200` makes readline insert wrong text.** The shim computes the longest common prefix from whatever it receives, and an LCP over a *subset* can be longer than the true LCP. On a lattice where the first 200 `Q*` matches all start with `QF`, typing `q<TAB>` inserts `QF` even though `QD…` exists past the cap. Better to suppress matches entirely (or raise the cap) when truncation occurs.

**7. Switch *arguments* are not skipped during word gathering.** `-switch` words are dropped but their values are not, so `show -write out.txt ele<TAB>` and `set -branch 2 ele<TAB>` both go dead. Either handle the value-taking switches or document the limitation.

**8. `c_f_pointer(line_c, line_p, [len(line_f)])` claims 4000 bytes of `rl_line_buffer`,** which is typically a much smaller heap allocation. The NUL always stops the loop so it works in practice, but it is a false array shape that ASAN will flag. Only text left of the cursor is used, so `[point+1]` is both safe and simpler.

## Nits

- `character(28) key` in `add_switch_matches` hardcodes the number rather than using `tao_switch_name_len`; gfortran warns on `key = 'show ' // key` (33→28) and `key = cmd_name` (40→28). Harmless today (longest real key is `show radiation_integrals`, 24 chars) but it is exactly the duplication this PR exists to remove. Nothing enforces the 28-char budget either — overflow means a silently truncated switch name or a missed lookup (see #4).
- `tao_switches_for` is called *inside* the parse loops, re-copying the array and re-scanning the 35-entry table on every switch token. Hoist it above the `do`.
- The `show merit` / `show top10` table rows are completion-only — those switches are parsed by `index()` at `tao_show_this.f90:3818` — but the module header's "Completion-only" list names only set/change/place. Please add them; drift prevention is the whole point.
- `tao_command_names` now drives `help <TAB>`, so the internal `debug`, `verbose`, and `tree` commands get advertised.
- `regression_tests/pytao/completion/test_pipe_complete.py` is not run anywhere: `pytao.yml` runs `./pytao/tests`, and `regression_tests/CMakeLists.txt` never references the pytao tree. `regression_tests/pytao/README.md`'s Organization list is also not updated.

## On the central question

I agree that duplicating the subcommand names is unavoidable here, and the "keep in sync with the select case" comments are the right mitigation. The refactor itself is faithful. My concerns are all about the new surface: the global readline state, the lost filename fallback, and the string-keyed lookup trading compile-time checking for silent runtime failure.

